PR #55, opened 2022-02-07
Fix mention of Matrix room in README

aaronraimist:
Matrix room aliases start with `#`


foolip (approved):

foolip:
Thanks @aaronraimist!

PR #102, opened 2022-08-03
Create the Interop Team Charter

foolip:
This will come into effect with https://github.com/web-platform-tests/rfcs/pull/120.


charter.md
```diff
@@ -1,29 +1,34 @@
...
 
-## Governance
+Browser features that don't affect web developers, like bookmarks, are out of scope.
+
+Writing specifications is not in scope, but the team can liaise with and work within other working groups to make progress in key areas.
 
-The team will meet regularly, with agenda issues created in the repo at least 48 hours in advance, and meeting minutes posted in the issue.
+## Governance
 
 The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
```
karlcow, line 20: What defines a participating organization?

Is it "track record of contributions"? https://github.com/web-platform-tests/interop/pull/102/files/274a676f9dbff051cea325f652e067821487f88b..0cee103e550562b2cfa44ab6d8788a9e0ad599e5#diff-d22b3d6550c34594a6e2d6b2148ed17907c5d598b10f407b6fbecf07bf25ea97R34

foolip: It's the list of organizations in https://github.com/web-platform-tests/interop/blob/main/supporters.md, who have in some sense "joined" the team. We should write down something about how to join, and there a "track record of contributions to the web platform" is the criteria we've discussed.

charter.md
```diff
@@ -0,0 +1,41 @@
...
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+The team is responsible for maintaining [Compat 2021](https://wpt.fyi/compat2021), [Interop 2022](https://wpt.fyi/interop-2022), and defining future interoperability efforts and metrics.
+
+## Scope
+
+The team works by prioritizing specific web platform features. The minimum requirements for a feature to be in scope are:
+
+- The feature is defined by a specification from the [IETF](https://www.ietf.org/), [Khronos Group](https://www.khronos.org/), [TC39](https://tc39.es/), [W3C](https://www.w3.org/), or [WHATWG](https://whatwg.org/). Other specifications can be considered on a case-by-case basis.
+- The feature can be tested in [web-platform-tests](https://github.com/web-platform-tests/wpt). Other test suites can be considered on a case-by-case basis.
```
foolip, line 12: I've filed https://github.com/web-platform-tests/interop/issues/117 about Test262, but it's not listed here now.

foolip:
@karlcow, @jensimmons mentioned you might be interested in reviewing this charter?

karlcow:
Yes it's on my todo for today.

charter.md
```diff
@@ -0,0 +1,41 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+The team is responsible for maintaining [Compat 2021](https://wpt.fyi/compat2021), [Interop 2022](https://wpt.fyi/interop-2022), and defining future interoperability efforts and metrics.
+
+## Scope
+
+The team works by prioritizing specific web platform features. The minimum requirements for a feature to be in scope are:
```
karlcow, line 9: As written, it looks a bit like the interop team is setting the agenda of what should be on the Web. It probably doesn't capture the reality of browsers implementations and the industrial priorities. Basically, the interop is here to help in improving the state of the web, but not to define what's next on the Web.


```suggestion
The team works by identifying specific web platform features which lack of interoperability among browser implementations **and** are in demand by web developers. 

The minimum requirements for a feature to be in scope are:
```

Probably something to add about limited volume of work to be successful. **TODO**

charter.md
```diff
@@ -0,0 +1,41 @@
...
+
+- The feature is defined by a specification from the [IETF](https://www.ietf.org/), [Khronos Group](https://www.khronos.org/), [TC39](https://tc39.es/), [W3C](https://www.w3.org/), or [WHATWG](https://whatwg.org/). Other specifications can be considered on a case-by-case basis.
```
karlcow, line 11: Specification is an umbrella term which covers many cases. Some of these are probably not acceptable for the work here. Example: A work proposal, or draft in a W3C WICG is not a good work item for the interop team. The goal is not to implement new things. The goal is to improve interoperability on things already implemented. 

```suggestion
- The feature is defined by a specification —on a standard track when the term exists— from the [IETF](https://www.ietf.org/), [Khronos Group](https://www.khronos.org/), [TC39](https://tc39.es/), [W3C](https://www.w3.org/), or [WHATWG](https://whatwg.org/). Other specifications can be considered on a case-by-case basis.  
**Example**: The [W3C Recommendation track](https://www.w3.org/2021/Process-20211102/#rec-track) is acceptable, but not the [W3C Note track](https://www.w3.org/2021/Process-20211102/#note-track)


- The feature is partially implemented in one or multiple browsers.
```

Issue: The term standard and specification means different things for different organizations. Some are super loosely defined. 

IETF is a good example of how difficult it can become… Anyone can propose an RFC as an [Internet Draft](https://www.ietf.org/how/ids/), and not all RFCs switch to the [standard track](https://www.rfc-editor.org/rfc/rfc2026.html#section-4).

[HTTP Semantics](https://datatracker.ietf.org/doc/rfc9110/) is a good candidate on the standard track, but [OAuth 2.0 for Browser-Based Apps](https://datatracker.ietf.org/doc/draft-ietf-oauth-browser-based-apps/) would not be.

**TODO**: Do we need a table for each organizations on what is standard material or not. -> #121

**TODO**: Should there be an implementation requirement? I added a third criteria. The reasons is that interop group is not here to do prototyping or even be a platform for starting new features even on the standard track. It's about improving the interop state of features which needs care and love to not be a painful point for webdevs. A missing webdev feature across the Web is not necessary an interop issue, it's just an idea which needs work in a standard group/interest group and be refined there. And hopefully if well develop it will never have to go through this group. -> #122

charter.md
karlcow, line 12:
```suggestion
- The feature can be **automatically** tested in [web-platform-tests](https://github.com/web-platform-tests/wpt). Other test suites can be considered on a case-by-case basis.
```

Added automatically. Many tests in web-platform-tests currently requires manual testing. We want to avoid this, to not burden ourselves into having to run manual tests. 

It's the second time that the charter introduces a case-by-case basis. While I understand that this is a way to give a bit of grease to the process, this is also an opportunity for having endless discussions. 

**TODO**: Open an issue about this case by case notion. What do we mean? What are the precedents? Was it needed? And Why? Was the group too constrained? -> #123

charter.md
```diff
@@ -0,0 +1,41 @@
...
+- The feature can be tested in [web-platform-tests](https://github.com/web-platform-tests/wpt). Other test suites can be considered on a case-by-case basis.
+
```
karlcow, line 13: Maybe we need an Out of scope section:

```suggestion
### Out Of Scope

Everything which is not defined in the scope section is out of scope.  It includes: 
```

charter.md
karlcow: Maybe we need room in the process on how to propose a new test suite, or new standard organization.

charter.md
```diff
@@ -0,0 +1,41 @@
...
+* Apple
+* Bocoup
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining the Team
+
+If your organization has a track record of contributions to the web platform you can request to join the Interop Team. Broad participation is welcome, but the existing team will make a decision based on consensus. There are also [other ways of contributing](./CONTRIBUTING.md).
```
karlcow, line 41:
```suggestion
If your organization has a track record of contributions to the implementation of web platform features, you may request to join the Interop Team. The decision will be made by the current [Team Members](#team-members) based on consensus.

You do not have be part of the interop Team to participate. There are also [other ways of contributing](./CONTRIBUTING.md).
```

How does an organization request to join? Open an issue? A Template? -> #124

karlcow (changes requested):
@foolip Thanks for the first draft. Super useful. Always easier to start commenting on an already drafted text.

I gave suggestions. Some of them might be up for discussions. I opened separate issues, but I might have missed some. Do not hesitate to ask me to open specific issues if you think my suggestions need more discussions.

charter.md
```diff
@@ -0,0 +1,41 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
```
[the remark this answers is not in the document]
foolip: I'm not so sure about this one, the list of members is _quite_ the same as the definition of the team, and more importantly people might follow the link and end up just scrolling the doc a bit. It's not a long charter :)

charter.md
foolip: So the intention here is to say that we aim to improve interoperability. But our approach isn't to just do it broadly and generically, but to prioritize specific web platform features.

I like your wording, but wonder if the last bit can be phrased more closely to the "prioritized by web developer and user needs" at the top? We could say that it causes problems for developers or users, perhaps.

charter.md
foolip: Thanks for filing those issues!

charter.md
foolip: I actually think we should leave the door open to manual tests, but we could leave that as one of the case-by-case situations. For example, if we found that OKLab gradients have some serious interop issue, but WPT is not yet able to test this for some reason, then having a few visual tests that we test once a month could be acceptable, if that's what makes the difference for improving interop.

charter.md
foolip: Good question!

I think we should also define that an organization can be removed by their own request, or by consensus to deal with the case that they're no longer around to request leaving.

charter.md
fantasai: +1 to @foolip on leaving things open to manual / non-WPT testing. If having the possibility truly ends up being a problem, you can restrict it then, but there are many things (e.g. Media Queries) that WPT is not currently good at testing that are nonetheless worth reaching interop on.

charter.md
fantasai: Proposed wording:
```
- The feature is defined by a standards-track specification from the following organizations:
      * [IETF](https://www.ietf.org/) (Standard Track)
      * [Khronos Group](https://www.khronos.org/)
      * [TC39](https://tc39.es/) (Stage 2+)
      * [W3C](https://www.w3.org/) (Recommendation Track)
      * [WHATWG](https://whatwg.org/)
   Other specifications can be considered on a case-by-case basis.
```

charter.md
gsnedders: I'd much rather we had _something_ in here about automated tests, and perhaps make it clear that the "case-by-case basis" includes manual tests in WPT?

foolip: How about this?

> The feature can be tested by automated tests in [web-platform-tests](https://github.com/web-platform-tests/wpt). Manual tests and other test suites can be considered on a case-by-case basis.

foolip: I've pushed this edit.

charter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The team is responsible for maintaining [Compat 2021](https://wpt.fyi/compat2021), [Interop 2022](https://wpt.fyi/interop-2022), and defining future interoperability efforts and metrics.
+
+## Scope
```
jgraham, line 7: IMO the "scope" section should be about the kind of work that the team can take on, not about the details of which proposals are expected to be accepted.

I'd suggest the scope as some thing like:

```
The output of the Interop project is metrics designed to measure and improve the interoperability of the web platform.

These metrics are expected to be comprised of:
* Groups of automated tests that can be used to automatically compute an interoperability score for specific platform features in implementations.
* Work to improve the interoperability of the platform that cannot be measured through test pass rates and is scored by consensus decision.

The Interop group takes ownership of the existing Compat-2021 and Interop-2022 projects, and those will form a basis for future work.

This is not a venue for standardisation, and work on new platform features is out of scope. However the group may liase with existing standards groups to help progress work which affects the interoperability of the platform, or the ability to test and measure interopability of existing features.
```

All the existing content about which proposals are actually likely to be accepted is really a consequence of the "proposals require consensus" rule, and therefore shouldn't form an independent part of the charter. Instead we should add some informative text to a proposer-facing README about the kinds of proposals that are likely to find consensus (i.e. those which are in a recognised standards group, are aimed at browsers, where all implementors are positive on the feature design, and where the level of stability of the specification is appropriate for implemention work to occur).

This has the advantage that we don't need to make charter-level changes if specific participants change their opinions about which kinds of proposals they are unwilling to accept.

charter.md
```diff
@@ -0,0 +1,41 @@
...
+
+The team works by prioritizing specific web platform features. The minimum requirements for a feature to be in scope are:
+
+- The feature is defined by a specification from the [IETF](https://www.ietf.org/), [Khronos Group](https://www.khronos.org/), [TC39](https://tc39.es/), [W3C](https://www.w3.org/), or [WHATWG](https://whatwg.org/). Other specifications can be considered on a case-by-case basis.
+- The feature can be tested by automated tests in [web-platform-tests](https://github.com/web-platform-tests/wpt). Manual tests and other test suites can be considered on a case-by-case basis.
+
+Browser features that don't affect web developers, like bookmarks, are out of scope.
+
+Writing specifications is not in scope, but the team can liaise with and work within other working groups to make progress in key areas.
+
```
jgraham, line 17: Not sure if this should be in "scope" or somewhere else, but possibly something like the following:

```
Initially the group will produce one set of interop metrics per year. This may be changed in the future by consensus decision.

If at any time the group is unable to reach consensus which features to include in its metrics, or the set of features which has consensus is considered insufficient to fulfill the goals of the group, the effort will be paused until the situation can be resolved.
```

(the last bit perhaps doesn't say anything new, but I want it to be clear that if e.g. we only get consensus on one feature we don't need to go ahead with producing an Interop metric from that, but can instead just say "this isn't working, we need to try something different")

charter.md
```diff
@@ -0,0 +1,41 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+The team is responsible for maintaining [Compat 2021](https://wpt.fyi/compat2021), [Interop 2022](https://wpt.fyi/interop-2022), and defining future interoperability efforts and metrics.
```
jgraham, line 5: (if my other suggestions are accepted) remove this, as it will be covered under "scope".

charter.md
```diff
@@ -0,0 +1,41 @@
...
+- The feature is defined by a specification from the [IETF](https://www.ietf.org/), [Khronos Group](https://www.khronos.org/), [TC39](https://tc39.es/), [W3C](https://www.w3.org/), or [WHATWG](https://whatwg.org/). Other specifications can be considered on a case-by-case basis.
+- The feature can be tested by automated tests in [web-platform-tests](https://github.com/web-platform-tests/wpt). Manual tests and other test suites can be considered on a case-by-case basis.
+
+Browser features that don't affect web developers, like bookmarks, are out of scope.
+
+Writing specifications is not in scope, but the team can liaise with and work within other working groups to make progress in key areas.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition.
```
jgraham, line 20: I understand the desire to have "scope" before "governance" but this feels like a key point that is needed to understand the rest of the document.

jgraham (changes requested):

charter.md
foolip: Sure, I'll reverse the order of the sections.

charter.md
foolip: I like the wording, and for lack of a better place put it at the end of the scope section.

charter.md
foolip: I kept this wording (with links) but put it in the scope instead.

charter.md
foolip: I've applied the suggestion with only minor edits/adaptations.

charter.md
foolip: I've added an "Out of Scope" with two paragraphs, but felt that "Everything which is not defined in the scope section is out of scope" wasn't necessary.

charter.md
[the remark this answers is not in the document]
foolip: Decision unless someone feels strongly: let's not link.

charter.md
foolip: I've updated this bit now. I think that "implementation of web platform features" was an improvement to what I had written originally, but I changed it to "contributions to the web-platform-tests project" to unambiguously fit all current members, and also it makes sense given the team is part of that project.

charter.md
foolip: This bit is gone now.

charter.md
foolip: Following the discussion in https://github.com/web-platform-tests/interop/issues/127#issuecomment-1234523551 I've removed details about how proposals are evaluated from the charter, and instead added it to https://github.com/web-platform-tests/interop/pull/115.

@fantasai I used your wording almost verbatim, since @jgraham indicated Mozilla would like a standards track expectation as well.

README.md
```diff
@@ -1,34 +1,7 @@
 # The Interop Project
 
-Welcome to the project Interop 2022 and Interop 2023!
+Welcome to the Interop project, an ongoing effort to make the web more interoperable in key areas, prioritized by web developer and user needs. This is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform. See our [charter](./charter.md) for more on our scope.
 
-This project is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), and exist to make the web more interoperable in key areas, prioritized by web developer and user needs. See our [charter](./charter.md) for more on our scope.
+**We are currently planning for Interop 2023 and welcome participation from anyone working on the web, please see the [Interop 2023 README](./2023/README.md) to learn more!**
```
karlcow, line 5: 👍

charter.md
```diff
@@ -0,0 +1,51 @@
...
+* Apple
+* Bocoup
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+If your organization has a track record of contributions to the web-platform-tests project, you may request to join the Interop Team by contacting interop@web-platform-tests.org. The decision will be made by the current team based on consensus.
```
jgraham, line 49: I think this should be rewritten in normative language e.g.

"""
The Interop team is made up of organisations with a track record of contributions to the web-platform-tests project.

Organisations my apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a consensus decision of existing members.

Existing members may leave the team at any time without further obligation.
"""

charter.md
```diff
@@ -0,0 +1,51 @@
...
+
+Team members can be removed by their own request or the consensus of the entire team.
```
jgraham, line 51: It's not quite clear what this means; if it's intended to cover disruptive members then presumably there wouldn't be consensus on removing them (because they would object). So the criteria has to be "consensus amongst the other team members" or something.

jgraham:
@tantek do you also want to review this charter proposal?

charter.md
foolip: It's in case someone just disappears.

jgraham: Well that still has the same issue; if someone disappears they can't consent to their own removal. I'm pretty sure the intent here is that members can be removed with the consent of the _rest_ of the team.

charter.md
```diff
@@ -46,6 +46,8 @@ The following organizations make up the Interop Team, in alphabetical order:
 
 ## Joining and Leaving the Team
 
-If your organization has a track record of contributions to the web-platform-tests project, you may request to join the Interop Team by contacting interop@web-platform-tests.org. The decision will be made by the current team based on consensus.
+The Interop Team is made up of organization with a track record of contributions to the web-platform-tests project.
 
-Team members can be removed by their own request or the consensus of the entire team.
+Organization my apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a consensus decision of existing members.
+
+Existing members may leave the team at any time without further obligation. If a member is no longer active in the team, they can be removed by consensus amongst the other team members.
```
jgraham, line 53: So as written, this doesn't handle the case where e.g. someone joins and then intentionally blocks all work, or they meet some threshold of being "active" but are not positively contributing to the goals of the project.

In practice I don't think that would happen, and if it did the other members would obviously find some workaround. But I'm still not sure it's worth adding in a conditional on the situations in which it's allowable to remove a member. If you really want a conditional it should at least be less restricitve than "no longer active".

foolip: This is a more antagonistic case than I'd considered. We could say that troublemakers can be ejected with the unanimous agreement of every other member, perhaps.

But maybe it's better to leave it to the wpt core team to sort out a situation like this? I'll make the edit to drop this bit.

charter.md
```diff
@@ -0,0 +1,59 @@
...
+
+The team is responsible for maintaining [Compat 2021](https://wpt.fyi/compat2021), [Interop 2022](https://wpt.fyi/interop-2022), and defining future interoperability efforts and metrics.
+
+Initially the team will produce one set of interop metrics per year. This may be changed in the future by consensus decision.
+
+If at any time the team is unable to reach consensus on which features to include in its metrics, or the set of features which has consensus is considered insufficient to fulfill the goals of the team, the effort will be paused until the situation can be resolved.
+
+### Extended Scope
+
+The team may also work on additional metrics or data visualizations to provide insights for a broader set of features, aimed at making the web more interoperable. For example, evolving the [Browser Specific Failures](https://github.com/web-platform-tests/results-analysis/blob/main/browser-specific-failures.js) metric is in this extended scope.
```
gsnedders, line 34: This whole section seems a bit redundant given the above Scope section says:

> The output of the Interop project is metrics designed to measure and improve the interoperability of the web platform.

So the Interop project's output and work is metrics and also additional metrics or data visualisations?

I don't see the point of this duplication.

And as discussed in the Interop meeting today, we shouldn't need any direct mention of BSF here—it is certainly a metric and thus falls under the scope. We should include it in the RFC, to make it explicit that the Core team is transferring it to the Interop team.

foolip: OK, I'll delete the section.

charter.md
bkardell: Given that we are in april 23, it seems weird to not go ahead and add 2023?

charter.md
```diff
@@ -0,0 +1,55 @@
...
+* Apple
+* Bocoup
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+The Interop Team is made up of organization with a track record of contributions to the web-platform-tests project.
```
bkardell, line 51: organizations*

karlcow (approved):
That looks good for me as it is now. 
I had forgotten we had not yet merged.

charter.md
foolip: Do you mean in the paragraph that starts "The team is responsible for maintaining"? I'll update it.

charter.md
foolip: Fixed, thanks!

charter.md
```diff
@@ -0,0 +1,55 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established.
```
boazsender, line 7: I think consensus usually means unanimous agreement. I recommend changing the definition to be unanimous agreement, or the label to be "rough consensus".

charter.md
```diff
@@ -0,0 +1,55 @@
...
+These metrics are expected to be comprised of:
+
+- Groups of automated tests that can be used to automatically compute an interoperability score for specific platform features in implementations.
+- Work to improve the interoperability of the platform that cannot be measured through test pass rates and is scored by consensus decision.
+
+The team is responsible for maintaining [Interop 2021](https://wpt.fyi/interop-2021), [2022](https://wpt.fyi/interop-2022), [2023](https://wpt.fyi/interop-2023), and defining future interoperability efforts and metrics.
+
+Initially the team will produce one set of interop metrics per year. This may be changed in the future by consensus decision.
+
+If at any time the team is unable to reach consensus on which features to include in its metrics, or the set of features which has consensus is considered insufficient to fulfill the goals of the team, the effort will be paused until the situation can be resolved.
```
boazsender, line 30: Not necessary to add anything here, but wondering if we have a sense of how we'll resolve situations like this. Will it be mediated? Or through a 1:1 meeting with disagreeing parties? If we do know, then it would be nice to add something about that, if we don't that's ok IMO.

boazsender:
This is really great @foolip! Just left a few comments/suggestions. No blockers from Bocoup. We support adopting this charter/merging this PR with or without additional changes.

charter.md
foolip: We haven't discussed how this might play out and I think it's hard to predict. I added this language to make it clear that the effort doesn't continue automatically in perpetuity, to recognize that if we can't reach consensus then the default outcome is doing nothing, i.e., pausing.

charter.md
```diff
@@ -0,0 +1,55 @@
...
+The output of the Interop project is metrics designed to measure and improve the interoperability of the web platform.
+
+These metrics are expected to be comprised of:
+
+- Groups of automated tests that can be used to automatically compute an interoperability score for specific platform features in implementations.
+- Work to improve the interoperability of the platform that cannot be measured through test pass rates and is scored by consensus decision.
+
+The team is responsible for maintaining [Interop 2021](https://wpt.fyi/interop-2021), [2022](https://wpt.fyi/interop-2022), [2023](https://wpt.fyi/interop-2023), and defining future interoperability efforts and metrics.
+
+Initially the team will produce one set of interop metrics per year. This may be changed in the future by consensus decision.
```
jgraham, line 28: I think we could remove this sentence? (I may have originally proposed it, not sure). It doesn't seem to add much.

charter.md
jgraham, line 30: In theory if we had >1 metric, the default would be to stop only those metrics for which we didn't have consensus. It's not really clear if this paragraph is intending to be specific to the case where there are zero agreeable things or not.

foolip: Since @boazsender also had questions about this paragraph, I'll just remove it instead of trying to spell out exactly the way things can eventually go wrong.

charter.md
boazsender: Sounds good to me

charter.md
```diff
@@ -25,8 +25,6 @@ These metrics are expected to be comprised of:
 
 The team is responsible for maintaining [Interop 2021](https://wpt.fyi/interop-2021), [2022](https://wpt.fyi/interop-2022), [2023](https://wpt.fyi/interop-2023), and defining future interoperability efforts and metrics.
 
-If at any time the team is unable to reach consensus on which features to include in its metrics, or the set of features which has consensus is considered insufficient to fulfill the goals of the team, the effort will be paused until the situation can be resolved.
```
jgraham: Rather than remove this entirely I would prefer to say something like:

In the event that the team is unable to reach sufficient consensus to fulfill the requirements of this charter (e.g. if it is unable to agree on any metrics to publish) the effort will be paused until the situation is resolved.

charter.md
```diff
@@ -0,0 +1,51 @@
...
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+The Interop Team is made up of organizations with a track record of contributions to the web-platform-tests project.
+
+Organization my apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a consensus decision of existing members.
```
zcorpan, line 49:
```suggestion
Organizations may apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a consensus decision of existing members.
```

karlcow:
I need to read this again. Since the last time I said ok.

charter.md
gsnedders: @foolip it doesn't seem like you ever did add it back in with your edits yesterday?

foolip: I didn't, because there was no agreement on what it should say. So I'm leaving it to you :)

charter.md
```diff
@@ -0,0 +1,51 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established.
+
+The chair of the team is responsible for organizing meetings, driving discussions to a conclusion, and recording decisions. The team appoints an individual as the chair based on consensus once per calendar year. There's no term limit; the same individual can be reappointed as chair.
```
karlcow, line 9: nit: Putting myself as a simple person outside, I was searching on https://web-platform-tests.org/admin/index.html and the github repo and was unable to answer the following questions:


* Who is the current chair?
* When is the rotation?
* Who were the previous chairs?

Maybe it doesn't matter. not sure. Maybe it could be written on the README.

foolip: We don't have a current chair because we don't have a charter, but @nairnandu has been acting in the role described here. I'd like to nominate @nairnandu for the chair position when the charter is accepted, but of course it's a group decision.

When decided, I think we should put the 6 participating organizations and the current chair in the top-level README.

charter.md
gsnedders: This was the main part of the discussion in https://github.com/web-platform-tests/interop/issues/329.

That said:

Most obviously, existing metrics will likely still need maintenance, even if we reach a point of deciding not to publish future ones.

It's also unclear what it means for it to be "paused". 

The motivation to have some text here I think is two-fold:

1. If at some point the team believes there is no need to publish further, future metrics (or at least not on an annual cadence), or
2. If at some point the team is unable to reach consensus on what a new metric should be.

In the former case I don't think we necessarily need to do anything (maybe maintenance could be handed over to the Core Team, but that's not necessarily the case).

In the latter case we could outline various processes, but I don't know it's worthwhile, and it's entirely plausible the group could still functionally maintain existing metrics.

charter.md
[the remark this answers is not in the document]
foolip: Sounds good on both accounts, can you propose text changes?

charter.md
```diff
@@ -0,0 +1,51 @@
...
+These metrics are expected to be comprised of:
+
+- Groups of automated tests that can be used to automatically compute an interoperability score for specific platform features in implementations.
+- Work to improve the interoperability of the platform that cannot be measured through test pass rates and is scored by consensus decision.
+
+The team is responsible for maintaining [Interop 2021](https://wpt.fyi/interop-2021), [2022](https://wpt.fyi/interop-2022), [2023](https://wpt.fyi/interop-2023), and defining future interoperability efforts and metrics.
+
+### Out of Scope
+
+This is not a venue for standardization, and work on new platform features is out of scope. However the group may liaise with existing standards groups to help progress work which affects the interoperability of the platform, or the ability to test and measure interoperability of existing features.
```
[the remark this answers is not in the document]
foolip: Is there a change to the text you'd like to see?

charter.md
foolip: @jgraham do you still want to keep some text here? I'm fine either way.

jgraham: I think it makes sense to say _something_ about what happens if we are unable to fulfill the requirements of the charter (e.g. because we can't reach consensus on continuing to produce metrics). I don't think it makes sense to try to enumerate every possible case.

Per @gsnedders points, I think "we continue to maintain existing Interop-style annual metrics but can't agree to publish new ones" would count as no longer fulfilling the requirements of the charter. I think the case where we do reach consensus, but the consensus is "we don't want to publish more metrics" can be understood as the group deciding to disband (that is, I don't think routine maintenance alone is enough to warrant the group existing). And I think the case where we settle on a single ongoing metric but with periodic updates to its composition counts as the group still operating as designed.

gsnedders: https://github.com/web-platform-tests/interop/pull/342 contains a proposal here.

charter.md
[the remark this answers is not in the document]
gsnedders: https://github.com/web-platform-tests/interop/pull/342 contains a proposal here.

charter.md
```diff
@@ -0,0 +1,51 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established.
```
dandclark, line 7: It's not defined whether consensus-breaking vetoes will be recorded in public notes or not. Unless we can agree to make this process fully public, let's be explicit that it's not. For example:

```suggestion
The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established. This process will be member-confidential; vetoes preventing consensus are not required to be made public.
```

Same goes for the votes for chair selection, and induction of new members.

charter.md
```diff
@@ -0,0 +1,51 @@
...
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+The Interop Team is made up of organizations with a track record of contributions to the web-platform-tests project.
+
+Organizations may apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a consensus decision of existing members.
```
dandclark, line 49: Unless we want to make this explicitly public (which would also be fine), we should be clear that it's not:

```suggestion
Organizations may apply to join the team by contacting interop@web-platform-tests.org. New members will be added with a member-confidential consensus decision of existing members.
```

dandclark (commented):
What IP policy does WPT operate under? Should this document state that that policy applies to contributions in Interop?

charter.md
foolip: I would be fine with having this be a normal (public) decision or a member-confidential one, with a preference for just making it public.

foolip:
> What IP policy does WPT operate under? Should this document state that that policy applies to contributions in Interop?

@dandclark there's no CLA, but we have https://github.com/web-platform-tests/wpt/blob/master/CONTRIBUTING.md and also lean on GitHub's ToS on [Contributions Under Repository License](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license) for the licensing.

I would expect all substantial contributions to go to either that repo, or via a standards group with its own IPR policy.

https://github.com/web-platform-tests/interop doesn't have a license. If you'd like us to add one, I'd suggest copying either https://github.com/whatwg/sg/blob/main/LICENSE or https://github.com/mdn/content/blob/main/LICENSE.md.

charter.md
foolip: I would also be OK with a simpler "everything is public" and no mention of member-confidential decisions in the charter.

dandclark:
> @dandclark there's no CLA, but we have https://github.com/web-platform-tests/wpt/blob/master/CONTRIBUTING.md and also lean on GitHub's ToS on [Contributions Under Repository License](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license) for the licensing.

Could we just have the charter link to these?

charter.md
jgraham: I suggest we note that the charter does not _require_ us to publish individual positions when forming consensus.

foolip: I think the distinction is whether the charter allows for _any_ decisions to be member-confidential, or if it allows for _some_ decisions to be member-confidential under certain circumstances. I would interpret "individual positions when forming consensus are not required to be made public" as the former. Is that the intention? Is there a downside to being public by default and requiring us to document where we allow for member-confidential decisions ahead of time? Clarity ahead of time is what we were missing in the Interop 2023 process, and what I would like to have.

karlcow: It's **not** about public vs member confidential. 
It's about diverse  opinions vs one voice. It's what we need to keep in mind in that discussion. 

What are the goals of interop and how do we best achieve these goals? The goal is to bring together everyone and show unity in an effort to remove implementation differences for web developers. 

The diversity of opinions in Web standards work is absolutely necessary and hence should be recorded in public, not as a tool for positions, but as a tool for diversity of technical opinions and how an idea grow into something, how the ideas evolve. 

I don't have the feeling this is the case for Interop work. This is a restricted group of organizations which are trying to define a series of actions to improve interoperability and make a better Web. It's a medical team around a sick body with one treatment.

foolip: @karlcow do you think the charter should say anything about how decisions are recorded? My original intention was to not mention it and handle it in the Interop 2024 process document, but since @dandclark raised the issue I think it's just as well we document it for clarity.

It seems reasonable to me to say that everything is public except where we've documented otherwise ahead of time, but if not, can someone suggest alternative wording?

foolip:
> > @dandclark there's no CLA, but we have https://github.com/web-platform-tests/wpt/blob/master/CONTRIBUTING.md and also lean on GitHub's ToS on [Contributions Under Repository License](https://docs.github.com/en/site-policy/github-terms/github-terms-of-service#6-contributions-under-repository-license) for the licensing.
> 
> Could we just have the charter link to these?

Would propose it as a new section, or where would we put it?

dandclark:
> Would propose it as a new section, or where would we put it?

Maybe just another line next to the CoC link, e.g.: https://github.com/web-platform-tests/interop/pull/351?

charter.md
karlcow: Let me come back to you a in a couple of weeks. Busy pipe this week.

jgraham: I honestly don't think the charter needs to say anything here, and to the extent it does it should note that it's not imposing any specific requirements. I think the Interop 2024 documentation should explicitly note that the outcomes of proposal selection will be communicated on behalf of the Interop group as a whole, since I think that led to good outcomes last year.

foolip: I would be OK with saying nothing here and sorting it out in the Interop 2024 process. I hope we can all agree that if the charter doesn't say, then it neither requires or rules out member-confidential decision making.

charter.md
foolip: I'm now leaning the other way, to document that this particular decision is member-confidential. @jgraham WDYT?

jgraham: As a point of pedantry (but this is a charter), the _decision_ will be public, the only question is whether any discussions or individual participant positions that go into forming that decision are required to be published. 

At least it seems unlikely that we'd want stronger publication requirements here than for any other work.

charter.md
foolip: @karlcow @gsnedders would saying nothing here be OK with you?

foolip:
Remaining blockers to finish this:
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]
- [ ]

charter.md
jgraham: My main concern is that such language might be wrongly taken to imply that other decision making processes are therefore not "allowed" to be member-confidential, by contrast. 

Also I don't know what the scope "the process" is here; would we be able to publicly minute the fact that FooCorp had applied to join if it was discussed in a meeting? In general does this imply that any and all mention of someone applying to join would be redacted from meeting minutes, or are you only intending it to cover the final details of the consensus-forming process (i.e. which participants objected, in the case the application was not accepted)?

foolip: What language would you suggest?

foolip: As for what the words should mean, we should respond in private if the request is private, in public if the request is public, and in either case we should have the option to just relay the outcome without justification or elaboration.

And I guess we don't want to set up that email address, so let's say you contact the chair instead.

jgraham: I don't think any change is needed (apart from fixing the mailing address).

charter.md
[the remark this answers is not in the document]
gsnedders:
```suggestion
This is not a venue for standardization work. Further, the team will avoid creating metrics which depend on significant standards work. However, this does not preclude the team from liaising with existing standards groups to help progress work which affects the interoperability of the platform, or the ability to test and measure interoperability of existing features (including features whose specification is determined to be ready-for-implementation but where there is yet to be any implementation).
```

charter.md
```diff
@@ -0,0 +1,55 @@
...
```
jgraham, line 53:
```suggestion
Organizations may apply to join the team by contacting the chair. New members will be added with a consensus decision of existing members.
```

charter.md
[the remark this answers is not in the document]
gsnedders: And that has now landed.

charter.md
gsnedders: And that has now landed.

charter.md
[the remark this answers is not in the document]
foolip: That SGTM, I'll apply the change.

foolip:
This charter looks good from Google's point of view.

charter.md
```diff
@@ -0,0 +1,55 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established.
+
+The chair of the team is responsible for organizing meetings, driving discussions to a conclusion, and recording decisions. The team appoints an individual as the chair based on consensus at a minimum once per calendar year. There's no term limit; the same individual can be reappointed as chair. If the team is unable to meet due to absence of a chair, the Core Team may appoint one directly.
```
bkardell, line 9: There's a couple of things about the chair... Its very vague.. i guess on purpose, but like.. Can we say how long a term is? Or that the team appoints for a term they specify, up to 1 year? I was thinking about co-chairs and interim chairs which are often helpful in working groups for very practical reasons and wondering if/how something like that would work.. could we say X is temporarily on leave, Y will serve as acting chair until they're back?

charter.md
```diff
@@ -0,0 +1,55 @@
...
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+The Interop Team is made up of organizations with a track record of contributions to the web-platform-tests project.
+
+Organizations may apply to join the team by contacting the chair. New members will be added with a consensus decision of existing members.
```
bkardell, line 53: Should there be a link to something in FAQ or something which will list the current chair?

charter.md
jgraham: I assume the "at minimum once per calendar year" means that you have to re-decide on the chair each year (i.e. the maximum duration of an appointment is a year) but you can do it more often than that (i.e. there could be a "temporary" chair under the process). I agree the outcome is somewhat implicit though.

dandclark:
I want to re-iterate our earlier concern: the charter should reflect that the group's decisions default to member-confidential if not otherwise specified. The fact that vetos are private and unattributed today is out of step with other web ecosystem governance practices and needs to be documented.

This need not limit anything about whether the group can decide to make certain decisions private or public, and we can craft the language to ensure that, but we should be very clear on what level of confidentiality there is for decisions if not otherwise specified.

charter.md
meyerweb: My interpretation agrees with James’.

nairnandu: Updated the text to clarify the yearly term for the chair

charter.md
nairnandu: We could add the details of the chair in the Interop readme page for a given year

charter.md
gsnedders: Per https://github.com/web-platform-tests/interop/pull/102#issuecomment-1623985431 it's clear that MS consider this still unresolved.

Unclear why this got resolved by @nairnandu?

chrishtr: @dandclark said at the meeting last week that he is ok with saying nothing for now, in the interest of landing a consensus document. That is why Nandu closed the comment.

dandclark: I suggested that as a possible outcome given lack of progress here but not quite ready to commit it yet, still need to confirm with some folks. Would much prefer to have a real resolution on this.

charter.md
dandclark, line 7: This is the language we'd like to propose, with the '+' line being my addition to the original text:

```suggestion
The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established. By default, this process will be public; votes of participating organizations may be shared outside the group. However, by consensus the group may choose in advance to make certain voting and decision-making processes member-confidential.
```

The goal here is to avoid uncertainty or surprises about what votes get will made public, without constraining the group from deciding to make parts of the process member-confidential.

meyerweb: This proposed language talks about processes which are not defined in the charter, such as voting, and which probably don’t belong in the charter in the first place, since processes should have the flexibility to be changed by consensus.

At any rate, those things are being worked out on the process side.  Statements about what processes or artifacts should or should not be confidential should be made on the process side, not in the charter.  Therefore, Brian and I believe this doesn’t belong in the charter.

foolip: @meyerweb I take "this process" in the suggestion to be the process of decision making. Maybe this wording would make it clearer?

> By default, decisions will be public; positions of participating organizations may be shared outside the team. However, by consensus the team may choose in advance to make certain decisions member-confidential.

@dandclark does that match your intended meaning?

Zooming out, I suggest that we decide on this charter wording and whether to make any part of Interop 2024 member-confidental at the same time. I could live with (approve and merge) the above wording and making Interop 2024 Focus Area Selection member-confidential.

foolip: After discussing and iterating today, I've updated the language to this:

> By default, the group will operate in public. However, specific decision-making processes can be made member-confidential by defining them as such in a process document published by the team.

I'll resolve this discussion, and lean on regular PR review to get that approved by everyone involved.

dandclark: Thanks @foolip , the updated language LGTM.

dandclark (approved):
Microsoft is good with the current state of the charter PR.

charter.md
```diff
@@ -0,0 +1,55 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established. By default, the group will operate in public. However, specific decision-making processes can be made member-confidential by defining them as such in a process document published by the team. Furthermore, process documents are only published with the consensus of the team.
```
foolip, line 7: Simplifying a little bit the edit I made during the late meeting:

```suggestion
The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established. By default, the group will operate in public. However, specific decision-making processes can be made member-confidential by defining them as such in a process document published with the consensus of the team.
```

charter.md
```diff
@@ -0,0 +1,55 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus, which is defined as support from at least two participating organizations and no opposition. Decisions can be made both in the repository and in meetings, as long as consensus can be established. By default, the group will operate in public. However, specific decision-making processes can be made member-confidential by defining them as such in a process document published with the consensus of the team.
```
gsnedders, line 7:
```suggestion
The team makes decisions based on consensus. A decision has consensus if it has support from at least two participating organizations and no opposition.

The team operates in public, except when otherwise allowed by this charter.

The team may publish a process document to define how it will develop a given metric. These process documents are adopted by a decision of the team. These process documents may define parts of their operation to be confidential to the team.
```

charter.md
```diff
@@ -0,0 +1,55 @@
...
+The Interop Team's scope includes metrics designed to measure and improve the interoperability of the web platform. These metrics are published once consensus to do so has been reached.
+
+These metrics are expected to be comprised of:
+
+- Groups of automated tests that can be used to automatically compute an interoperability score for specific platform features in implementations.
+- Work to improve the interoperability of the platform that cannot be measured through test pass rates and is scored by consensus decision.
+
+The team is responsible for maintaining [Interop 2021](https://wpt.fyi/interop-2021), [2022](https://wpt.fyi/interop-2022), [2023](https://wpt.fyi/interop-2023), and defining future interoperability efforts and metrics.
+
+If the Team is unable reach consensus to publish any new metrics, and has been unable to reach consensus for three months, the Team's functions revert to the Core Team and this charter is annulled.
```
gsnedders, line 30:
```suggestion
If the team is unable reach consensus to publish any new metrics, and has been unable to reach consensus for three months, the team's functions revert to the Core Team and this charter is annulled.
```

We otherwise use "Interop Team" and "team", thus be consistent and use lowercase here.

charter.md
```diff
@@ -0,0 +1,55 @@
...
+* Google
+* Igalia
+* Microsoft
+* Mozilla
+
+## Joining and Leaving the Team
+
+The Interop Team is made up of organizations with a track record of contributions to the web-platform-tests project.
+
+Organizations may apply to join the team by contacting the chair. New members will be added with a consensus decision of existing members.
```
gsnedders, line 53: There was some earlier discussion about whether we want the details of any discussions and/or vote to determine consensus to be public here. I think we merged it into the broader public/private discussion, but it may be worthwhile checking we are all okay with this being public, which per the current draft it would be.

charter.md
```diff
@@ -0,0 +1,55 @@
...
+
+The chair of the team is responsible for organizing meetings, driving discussions to a conclusion, and recording decisions. The team appoints an individual as the chair based on consensus at a minimum once per calendar year. The term limit for the chair will be 1 year and the same individual can be reappointed as chair, if there is consensus. If the team is unable to meet due to absence of a chair, the Core Team may appoint one directly.
+
+Meeting time and agenda should be announced by the chair at least 48 hours in advance, and meeting minutes posted in the repository.
+
+The Interop Team's work is subject to the [WPT code of conduct](https://github.com/web-platform-tests/wpt/blob/master/CODE_OF_CONDUCT.md).
+
+All contributions are licensed under the terms in [CONTRIBUTING.md](https://github.com/web-platform-tests/wpt/blob/master/CONTRIBUTING.md).
+
+Substantive changes to this charter are done via the [web-platform-tests RFC process](https://github.com/web-platform-tests/rfcs).
```
gsnedders, line 17:
```suggestion
Substantive changes to this charter are done via the [web-platform-tests RFC process](https://github.com/web-platform-tests/rfcs). Existing members of the team are notified that when such an RFC is created.
```

Do we believe the one-week rule for RFCs suffices? When the team has previously had biweekly meetings it was at times the case people didn't look at issues more often than the meetings, so an RFC could be filed and approved between two meetings without all those active in the Interop Team being aware of it.

This is also notably different to the W3C where re-chartering (including any substantive changes) requires rejoining the group—it seems undesirable that we might end up with members of the group who don't agree with the charter after a substantive change has been made.

jgraham: I think I agree in general that the RFC period as specified is rather short for cases where there's non-trivial disagreement and that the process we use in practice allows more time to provide feedback. The escalation process as written has never been used, and it's clear that in cases like this it would not actually work.

Maybe we should also look to update that to reflect reality better.

foolip: I think we'd need to change the RFC process to fix this, and this charter can't add rules to the RFC process. What I'd like to happen in practice is that changes are run by the Interop Team and that we're given at least 2 weeks to react.

I'm happy to just apply the suggested working, but @gsnedders @jgraham would that be good enough for you?

charter.md
foolip: How about like this?

```suggestion
Organizations may apply to join the team by contacting the chair. New members will be added with a consensus decision of existing members. The deliberations are confidential to the existing members.
```

gsnedders: Seems fine to me.

foolip: Thanks, I'll apply the change.

charter.md
foolip: Here's a wording change I think does the job:

```suggestion
Substantive changes to this charter require the consensus of the Interop Team and should go through the [web-platform-tests RFC process](https://github.com/web-platform-tests/rfcs).
```

In practice, the Interop Team would first internally agree to the change, and then put it through the RFC process.

jgraham: I don't think it can use 'should'. Maybe: "Substantive changes to this charter can be made using the wpt RFC process, with an additional requirement of explicit consensus from the Interop Team participants."

charter.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus. A decision has consensus if it has support from at least two participating organizations and no opposition.
+
+The team operates in public, except when otherwise allowed by this charter.
+
+The team may publish a process document to define how it will develop a given metric. These process documents are adopted by a decision of the team. These process documents may define parts of their operation to be confidential to the team.
```
jgraham, line 11: I think this "may" needs to be a "must".

charter.md
```diff
@@ -0,0 +1,59 @@
+# Interop Team Charter
+
+The Interop Team aims to make the web more interoperable in key areas, prioritized by web developer and user needs. This team is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+
+## Governance
+
+The team makes decisions based on consensus. A decision has consensus if it has support from at least two participating organizations and no opposition.
+
+The team operates in public, except when otherwise allowed by this charter.
```
jgraham, line 9: I assume the intent is that this applies transitively to the process documents defined below, but for the avoidance of doubt, we should say "by this charter, or by process documents specific to given metrics or other deliverables of the team."

jgraham (changes requested):

charter.md
foolip: Alright, so:

```suggestion
Substantive changes to this charter are done via the [web-platform-tests RFC process](https://github.com/web-platform-tests/rfcs). Substantive changes to this charter can be made using the wpt RFC process, with an additional requirement of explicit consensus from the Interop Team participants.
```

charter.md
foolip:
```suggestion
The team must publish a process document to define how it will develop a given metric. These process documents are adopted by a decision of the team. These process documents may define parts of their operation to be confidential to the team.
```

charter.md
foolip:
```suggestion
The team operates in public, except when otherwise allowed by this charter, or by process documents specific to given metrics or other deliverables of the team.
```

gsnedders (approved):

jgraham (approved):

PR #111, opened 2022-08-18
Copy the 2022 RFC into this repo

foolip:


foolip:
The idea here is that we'll have a 2023/ directory, and it's convenient to 2022/ in the same repo.

jgraham (approved):

zcorpan (approved):

PR #130, opened 2022-09-14
Add an Investigation Effort Proposal template

foolip:
This completes the public documentation for Interop 2023 proposals.

### DIFF
--- a/.github/ISSUE_TEMPLATE/focus-area-proposal.yml
+++ b/.github/ISSUE_TEMPLATE/focus-area-proposal.yml
@@ -9,7 +9,7 @@ body:
 
       Please see the [Interop 2023 README](https://github.com/web-platform-tests/interop/blob/main/2023/README.md) for the timeline and what to expect from the process.
 
-      If this feature does not have a high quality specification and tests, an [Investigation Effort](https://github.com/web-platform-tests/interop/issues/new?template=investigation-effort-proposal.yml) may be more appropriate.
+      If this feature is widely implemented, but lacks a high quality specification or tests, an [Investigation Effort](https://github.com/web-platform-tests/interop/issues/new?template=investigation-effort-proposal.yml) may be more appropriate.
 - type: textarea
   attributes:
     label: Description
--- a/.github/ISSUE_TEMPLATE/focus-area-proposal.yml
+++ b/.github/ISSUE_TEMPLATE/focus-area-proposal.yml
@@ -1,0 +1,44 @@
+name: Focus Area Proposal
+description: Proposal for an Interop 2023 Focus Area
+labels: focus-area-proposal
+body:
+- type: markdown
+  attributes:
+    value: |
+      Thanks for taking the time to make a proposal for Interop 2023!
+
+      Please see the [Interop 2023 README](https://github.com/web-platform-tests/interop/blob/main/2023/README.md) for the timeline and what to expect from the process.
+
+      If this feature does not have a high quality specification and tests, an [Investigation Effort](https://github.com/web-platform-tests/interop/issues/new?template=investigation-effort-proposal.yml) may be more appropriate.
+- type: textarea
+  attributes:
+    label: Description
+    description: Brief description of the feature. Try to be clear about what parts of the feature are included and not.
+- type: textarea
+  attributes:
+    label: Rationale
+    description: |
+      Reasons for including the feature. A few dimensions to consider are:
+
+      - Web developer sentiment from polls or surveys, for example, [MDN surveys](https://insights.developer.mozilla.org/) or [State of CSS](https://stateofcss.com/).
+      - Examples of sites or libraries working around a missing feature or browser bug.
+      - Browser bugs or browser bug statistics, for example, [Chromium](https://bugs.chromium.org/), [Gecko](https://bugzilla.mozilla.org/),  or [WebKit](https://bugs.webkit.org/).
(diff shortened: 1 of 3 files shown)


rachelandrew (commented):
This may have been discussed elsewhere that I've missed, but I'm concerned asking people to identify whether a feature has "high quality specifications and tests" is going to act as a barrier to developers submitting features. How would a regular webdev know if something has those? It feels as if deciding whether something is a focus area or area of investigation is something we are in a good place to decide, from the community we just want to know what they are struggling with.

.github/ISSUE_TEMPLATE/focus-area-proposal.yml
```diff
@@ -1,19 +1,29 @@
...
 - type: markdown
   attributes:
     value: |
-      **Interop 2023 has not yet launched and this documentation is still being finalized. Please do not submit proposals yet.**
- 
       Thanks for taking the time to make a proposal for Interop 2023!
 
       Please see the [Interop 2023 README](https://github.com/web-platform-tests/interop/blob/main/2023/README.md) for the timeline and what to expect from the process.
+
+      If this feature does not have a high quality specification and tests, an [Investigation Effort](https://github.com/web-platform-tests/interop/issues/new?template=investigation-effort-proposal.yml) may be more appropriate.
```
jgraham, line 12: This sounds like something which is early in the specification process might be suitable for an Investigate effort. We probably want to say "is already widely implemented, but lacks a high quality specification or tests" or similar.

2023/README.md
```diff
@@ -15,19 +13,20 @@ If you would like to make a proposal or contribute in other ways, read on.
...
-
 If you've had problems using a feature on the web because of differences between browsers, or because it isn't implemented in all browsers, it may be a good proposal for Interop 2023!
 
 Before making a proposal, here's what to expect:
 
-- Only features which have high quality specifications and tests are in scope. Interop 2023 is not a venue for specifying new features; that work happens in working groups within organizations such as W3C and WHATWG.
+- There are two kinds of proposals:
+  - **Focus Area**: Most proposals should fall into the category, where the goal is for browsers to pass a set of tests for a feature. Only features which have high quality specifications and tests are in scope.
+  - **Investigation Effort**: In cases where there aren't high quality specifications and tests, an investigation effort may be appropriate. An investigation efforts consists of a set of tasks that will bring the feature up to the bar required for a Focus Area in the future. If this involves any standards work, that work must be done in the appropriate standards group, and those doing the work need to join those groups.
```
jgraham, line 22: Again, I think we should make it clearer that Investigate efforts are for areas with known interop problems, not a venue for progressing early-stage standards work.

.github/ISSUE_TEMPLATE/focus-area-proposal.yml
foolip: I took that wording almost verbatim.

2023/README.md
foolip: Also updated.

foolip:
@rachelandrew I agree this is hard to judge, but I think we do need to make investigation efforts explicit, since we want them to be well defined before we accept them into Interop 2023. This year, we went ahead with 3 investigation efforts that weren't clear at the outset, and we're deciding as we go what the precise goals are and how we'll score them.

I've tried to make Focus Area the clear default, and even linked to it from the Investigation Effort template, but is there more we could do?

rachelandrew:
@foolip I think what you are aiming for is correct, we want to make sure the investigation efforts are well defined, I just think that's for us to figure out, not a web developer who has the problem. I don't want the result we are aiming for to gatekeep initial suggestions, if that makes sense?

foolip:
@rachelandrew I think the approach you're suggesting is that we have just one proposal template (like before this PR) and that we figure out along the way what is going to be a focus area and what's an investigation effort. I think that could work, but in a previous meeting I promised to split the templates and try to steer people toward Focus Area, and there seemed to be agreement on that approach.

@jgraham do you want to weigh in with your preference on this?

jgraham:
I think it makes sense to split the template. To some extent if you don't know enough to know whether a feature is well specified and has some tests, we probably want to instead encourage you to find someone to help you submit the proposal, because the effort required to make such proposals actionable after submission is likely to be very high. If we have specific volunteers to help mentor web developers who want to submit proposals but don't have the background to know the spec/testing situation maybe we could create a matric channel where those volunteers could hang out, and point prospective proposers there for help in constructing a proposal.

zcorpan:
I don't mind having support for proposal writers in Matrix, but another option we could consider is to allow for incomplete proposals and iron out the missing details in the issue comments and edit the OP as appropriate. We could use labels to indicate what is needed for the proposal (e.g. a list of tests or review spec quality).

jgraham:
I think we will inevitably allow propsals to mature after initial submission. But I think there's a meaningful difference between a proposal where the submitter has a good understanding of what we're looking for and just needs additional time to verify which tests are relevant &c. and a proposal where the proposer doesn't even know if the feature in question is well specified or not.

jgraham (approved):
I think this looks broadly fine; if we start getting proposals that don't fit the scope (e.g. "Investigate" proposals for incubation-stage ideas) we can make further tweaks to the templates / README.

foolip:
I'm confident this is broadly in line with what we've discussed in past meetings, so I'll go ahead and merge so that it's easier to review what the actual issue templates look like, and to spot broken links and such.

karlcow:
late to the party. The discussion here makes sense. 
There are two levels where we want to be tolerant in what we receive, but strict in what we work with for the future. 👍

PR #288, opened 2023-02-16
Create a workflow for maintaining a list of labeled tests in-repo

foolip:


.github/workflows/labels.yml
```diff
@@ -0,0 +1,25 @@
+name: labels
+on:
+  # Trigger every Monday morning, or manually.
+  schedule:
+    - cron: 55 5 * * 1
```
stof, line 5: Scheduled workflows are disabled by Github after 60 days without activity on the repository. will this repo have sufficient activity so that this does not happen or should the scheduled workflow be in the wpt repo, using a workflow dispatch to trigger the workflow on this repo ?
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64327 and https://github.com/DefinitelyTyped/dt-mergebot/pull/457

Note that instead of a personal access token with a `repo` scope (which grants full access to the repo) stored as a secret, it might be possible to define a GitHub App installed in this interop repo and having the `actions:write` permission. The wpt repo would then have a secret containing the id and secret key of that app to be able to get a temporary token to authenticate as that app using https://github.com/marketplace/actions/create-github-app-token (listing the interop repo as part of the repositories accessible by that token).

PR #390, opened 2023-08-08
Add a proposal for the Interop 2024 process

gsnedders:
To set out the goals behind this:

* Create a clear process that’s easy to execute
* Invite proposals from anyone interested, with clear documentation of the project’s purpose and process — so potential proposers understand the project, its values, and the requirements for successful proposals
* Evaluate and assess proposals in a realistic and doable fashion, greatly reducing the amount of time needed to do so
* Reach a result where the Focus Areas and Investigations chosen support the mission of the Interop Project

---

This is long, and clearly not totally complete—but we think this serves as a good starting point for future discussion. We think that setting out, explicitly, the scope and the goals for the Interop Project (which is—really—out of scope of the broader Interop Team charter, as the Interop Team will own things above and beyond the project) serves as a good introduction for what the process should achieve.

### DIFF
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -6,7 +6,7 @@
 
 1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
 2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
-3. Elimination Round 1: Disqualification — one week (Oct 12 – Oct 19)
+3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
 4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
 5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -2,7 +2,7 @@
 
 ## Process
 
-### Timeline — 10 weeks total
+### Timeline
 
 1. Call for Proposal Period — three weeks (Sep 14 – Oct 5)
 2. Continued discussion and refinement of proposals — one week (Oct 5 – Oct 12)
@@ -10,7 +10,7 @@
 4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
 5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
 
-
+Interop 2024 and the set of accepted proposals will be announced in January 2024.
 
 ### Call for Proposal Period
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -1,58 +1,5 @@
 # Proposal for Interop 2024 Process
 
-## The Purpose of the Interop Project
-
-*Improve interoperability significantly for the benefit of web developers and users.*
-
-The goal of the Interop Project is to improve the web by making it easier to make websites and web apps that work in every browser or browser engine at the same time.
-
-This is done by increasing the amount of “interoperability” between browsers — when each browser engine has implemented the same technology the exact same way, as bug-free as possible.
-
-Today’s browsers have made a commitment to implement web technology according to a shared web standard, created in organizations such as the W3C or WHATWG, where technologies such as CSS and HTML are officially defined.
-
-There is a seemingly infinite amount of work that browser engineering teams could be focused on. The Interop Project provides incentives to focus on the specific and practical work that will help web developers most in the coming year.
-
-## Scope
-
-The Interop Project is a collaboration between organizations that implement web technology in browser engines. It’s a shared agreement to prioritize work on certain web technologies to improve interoperability. And to keep track of that work by using automated tests to score how much progress each participating browser has made reaching the shared goals.
-
-There are many fantastic ideas for what the web could become. But the Interop Project is not the place to begin making those dreams a reality. This is a place to focus on interoperable implementations of ideas that have already been defined in detail in web standards. And it’s a place to focus on interoperability through the use of automated tests to evaluate whether or not a particular browser matches what the web standard says it should be doing. In order for this to be possible, the Interop Project only focuses on specific kinds of web technology.
-
-Everything chosen for the next year’s project must be defined in a sufficiently mature standards-track web specification. If there is no web standard, then there is no definition of what a particular technology “should” be doing. The Interop Project is not an organization for defining web standards. That work is important, and must happen in the appropriate bodies, not here.
-
-Everything chosen for the next year’s project must be testable using automated tests on existing testing platforms. A proposal cannot be accepted if there are no automated tests to evaluate conformance to the web standard; if it’s not possible to test the technology being proposed; or if the tests would need to run in environments that are not yet available on the infrastructure for Web Platform Tests (for example, in 2022 and 2023, testing on mobile operating systems was not possible).
-
-If there is a particular blocker making it hard to test a certain feature, that could make for a good Investigation Project proposal. And ideas about how the WPT/Interop Project testing infrastructure can be improved for the future can be proposed as Investigation Projects, to be considered as work for the Interop Project to take on in the next year.
-
-### **Requirements for Focus Area proposals**
-
-* This feature has a mature-enough web standard from an established standards development organization such as IETF, Khronos Group, TC39, WHATWG, or the W3C.
-* This feature has high quality tests in WPT or Test262.
-
-### **Guidance for Prioritizing Focus Area proposals**
-
-We should prioritize areas where web developers have had problems using them because of differences between browsers, or because they aren’t implemented in all browsers. There are many sources for such information:
-
-* Perhaps there are issues filed about such problems in the Chromium, Gecko, or WebKit bug trackers.
-* Perhaps there are many complaints about this feature in public discussions on sites like Stack Overflow.
-* Perhaps this feature ranks in high-demand or as a pain-point consistently in polls and surveys of web developers.
-* Perhaps there are documented examples of sites or libraries working around the fact that it is missing or not interoperable.
-
-When assessing the prioritization of existing web technology, it can be helpful to assess how often the technology is being used currently. (Such thinking should also consider important technology might not be in popular use if a lack of interoperability is a painful barrier.)
-
-* The feature has high usage via Chrome use counters.
-* The feature has high usage via HTTP Archive.
-* The feature has a high number of mentions on developer resources like Stack Overflow.
-
-There are also guiding values that should be taken into consideration, beyond pure demand:
-
-* This feature has a positive impact on accessibility.
-* This feature has a positive impact on internationalization.
-* This feature has a positive impact on privacy.
-* This feature has a positive impact on security.
-
-
-
 ## Process
 
 ### Timeline — 10 weeks total
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -49,7 +49,7 @@ A fourth week will allow for continued discussion and refinement of already-subm
 
 ### Elimination Round 1 : Disqualification
 
-After the Call of Proposals Period has ended, the organizing committee will finalize an official list of all Focus Area proposals.
+After the Call of Proposals Period has ended, the Interop Team will finalize an official list of all Focus Area proposals.
 
 Any Focus Area proposal that does not meet the minimum publicly-stated selection criteria will be considered for preliminary elimination. For example:
 
@@ -96,7 +96,7 @@ NOTE: in the past, “Exclude” was a way to express many different kinds of se
 
 ## Interop 2023 Evaluation
 
-During the same period organizations are making Round 3 decisions, the organizing committee should evaluate the status of Interop 2023, and decide which Focus Areas will be retired, and which will be carried over.
+During the same period organizations are making Round 3 decisions, the Interop Team should evaluate the status of Interop 2023, and decide which Focus Areas will be retired, and which will be carried over.
 
 \[XXX]
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -24,7 +24,7 @@ An open call for proposals will begin on a specific date, and run for three week
 4. Focus Area proposals must:
     1. Name a specific web technology.
         1. This can be a web technology that has not yet been implemented in any or all browser engines.
-        2. Or this can be a web technology that has already been implemented in all three engines, but those implementations have multiple test failures that reveal a lack of compliance to the web standard, causing a lack of interoperability.
+        2. Or this can be a web technology that has already been implemented in all three engines, but those implementations have test failures that reveal a lack of compliance to the web standard, causing a lack of interoperability.
     2. The feature must be defined in a sufficiently mature standards-track web specification from one the following organizations. Please provide link to the web standard.
         1. [IETF](https://www.ietf.org/) (Standard Track)
         2. [Khronos Group](https://www.khronos.org/)
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -24,7 +24,7 @@ An open call for proposals will begin on a specific date, and run for three week
 2. Proposals should be for either:
     1. Focus Area, or
     2. Investigation Project
-3. Related proposals should be filed separately. For example, all the ideas for “Web Compat” should be filed as a set of separate issues, not one issue. This makes it easier to discuss and prioritize each separately. Bundles of proposals risk being rejected or deprioritized.
+3. Related proposals should be filed separately. For example, a group of specific features to address under “Typography” should be filed as separate issues, not one issue. This makes it easier to discuss and prioritize each separately. Bundles of proposals risk being rejected or deprioritized.
 4. Focus Area proposals must:
     1. Name a specific web technology.
         1. This can be a web technology that has not yet been implemented in any or all browser engines.
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -88,7 +88,7 @@ For example, if there are 20 remaining proposals, each organization will choose 
 
 (If there appear to be quite a few proposals remaining in this round, the group can decide to only ask for each organization to submit their Top `X` number, while rating the rest zero. For example, if 50 proposals remain at this point, it likely makes sense to only submit a Top 25, and to not worry about ranking 26-50 in a specific order.)
 
-After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom.
+After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom. The team will decide by consensus which of the proposals will be included in Interop 2024 and how they will be grouped into focus areas and investigation efforts.
 
 NOTE: in the past, “Exclude” was a way to express many different kinds of sentiments — including: does not meet the minimum qualifications, or this is low priority to us. We expect “veto” to be used much less than “exclude”, because there are other ways to express those sentiments. Vetoing is specifically for objecting to a particular proposal being included *at all,* no matter how highly ranked other organizations might find it. It’s for expressing objection to a particular web technology, perhaps to something an organization has already expressed an objection to in the web standards process or otherwise has an objection to implementing.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -42,7 +42,7 @@ As soon as each proposal is submitted as a GitHub issue, discussion of that prop
 
 At the end of the three weeks, the deadline will end the period when brand new proposals can be submitted.
 
-A fourth week will allow for continued discussion and refinement. The deadline at the end of the fourth week ends the opportunity to significantly change the scope of the proposal.  The proposal will be evaluated on its state at that time.
+A fourth week will allow for continued discussion and refinement of already-submitted proposals. The deadline at the end of the fourth week ends the opportunity to significantly change the scope of proposals.  The proposals will be evaluated on their state at that time.
 
 
 ## Focus Area Selection
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -82,7 +82,7 @@ In this round, each organization ranks the proposals in order of importance from
 
 Any proposal that gets a veto is immediately eliminated.
 
-Each proposal gets assigned points from each organization, according to their rank, and an ordered list of all remaining proposals is created. Where `X` is number of proposals that are ranked, and `Y` is the ranking a particular proposal gets from a certain organization, then that proposal gets `X-Y`  points from that organization. Once all 6 rankings are in for that proposal, the total points from all six organizations is known. If that proposal is not eliminated by a veto, then it has a certain number of points, and can be listed in order among the other proposals.
+Each proposal gets assigned points from each organization, according to their rank, and an ordered list of all remaining proposals is created. Where `X` is number of proposals that are ranked, and `Y` is the ranking a particular proposal gets from a certain organization, then that proposal gets `X-Y+1`  points from that organization. Once all 6 rankings are in for that proposal, the total points from all six organizations is known. If that proposal is not eliminated by a veto, then it has a certain number of points, and can be listed in order among the other proposals.
 
 For example, if there are 20 remaining proposals, each organization will choose a number 1, number 2, down to number 20. Let’s say a proposal for flying cars gets the following ranks: 1, 5, 7, 10, 14, 20. That will give it the following points: 20, 15, 13, 10, 6, 1 = 65 points total.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -57,7 +57,7 @@ Any Focus Area proposal that does not meet the minimum publicly-stated selection
 2. The proposed technology is already interoperable according to WPT results.
 3. The proposed technology is not on a standards track, or its standard is not sufficiently mature.
 
-Decisions will be made at a specific meeting for discussion of preliminary eliminations. Any member organization can bring up a proposal they believe should be eliminated. Brief debate can be had. If **there is quick consensus**, that proposal will be eliminated. If there is not consensus and a longer debate is needed, the proposal will be not be eliminated and will instead move on to the next round.
+Decisions will be made at a specific meeting for discussion of preliminary eliminations. The minutes of this meeting are confidential to the team. Any member organization can bring up a proposal they believe should be eliminated. Brief debate can be had. If **there is quick consensus**, that proposal will be eliminated. If there is not consensus and a longer debate is needed, the proposal will be not be eliminated and will instead move on to the next round.
 
 1. Setup *Disqualification meeting*
 2. Each organization can propose items for disqualification.
@@ -70,7 +70,7 @@ Decisions will be made at a specific meeting for discussion of preliminary elimi
 
 At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
 
-Each member organization chooses 40 proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
+Each member organization chooses 40 proposals they would like to see move on to Round 3; these selections are confidential to the team. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
 
 The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
 
@@ -78,7 +78,7 @@ The goal is to find a balance so all of the proposals that are most likely to be
 
 The official list of Focus Area proposals is updated to reflect the elimination of proposals not chosen by any organization. Each organization now has 20 days to review the list of remaining proposals and make their Round 3 decisions.
 
-In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise.
+In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise. The entire contents of this round is confidential to the team.
 
 Any proposal that gets a veto is immediately eliminated.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -80,7 +80,7 @@ The official list of Focus Area proposals is updated to reflect the elimination 
 
 In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise. The entire contents of this round is confidential to the team.
 
-Any proposal that gets a veto is immediately eliminated.
+Any proposal that has an objection is immediately eliminated.
 
 Each proposal gets assigned points from each organization, according to their rank, and an ordered list of all remaining proposals is created. Where `X` is number of proposals that are ranked, and `Y` is the ranking a particular proposal gets from a certain organization, then that proposal gets `X-Y+1`  points from that organization. Once all 6 rankings are in for that proposal, the total points from all six organizations is known. If that proposal is not eliminated by a veto, then it has a certain number of points, and can be listed in order among the other proposals.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -5,7 +5,7 @@
 ### Timeline
 
 1. Call for Proposal Period — three weeks (Sep 14 – Oct 5)
-2. Continued discussion and refinement of proposals — one week (Oct 5 – Oct 12)
+2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
 3. Elimination Round 1: Disqualification — one week (Oct 12 – Oct 19)
 4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
 5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -70,11 +70,9 @@ Decisions will be made at a specific meeting for discussion of preliminary elimi
 
 At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
 
-Each member organization chooses `N` number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
+Each member organization chooses 40 number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
 
-For example, let’s imagine there are 112 focus area proposals submitted. Let’s set `N` to 10 — so each of the six organizations gets to choose 10 proposals. There will likely be significant overlap, but not complete. So imagine 32 proposals get chosen by at least one organization. (It will be no less than 10, and no more than 60.) This means the remaining 80 proposals are eliminated.
-
-If `N` is increased, then more proposals will make it into the next round. The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
+The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
 
 ### Elimination Round 3: Prioritization
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -7,7 +7,7 @@
 1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
 2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
 3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
-4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
+4. Elimination Round 2: Reduction — two weeks (Oct 23 – Nov 3)
 5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
 
 Interop 2024 and the set of accepted proposals will be announced in January 2024.
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -12,6 +12,8 @@
 
 Interop 2024 and the set of accepted proposals will be announced in January 2024.
 
+(This timeline can be adjusted as needed, at any time, by consensus of the Interop Team.) 
+
 ### Call for Proposal Period
 
 An open call for proposals will begin on a specific date, and run for three weeks.
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -47,7 +47,7 @@ A fourth week will allow for continued discussion and refinement of already-subm
 
 ## Focus Area Selection
 
-### Elimination Round 1 : Disqualification
+### Elimination Round 1: Disqualification
 
 After the Call of Proposals Period has ended, the Interop Team will finalize an official list of all Focus Area proposals.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -14,6 +14,8 @@ Interop 2024 and the set of accepted proposals will be announced in January 2024
 
 (This timeline can be adjusted as needed, at any time, by consensus of the Interop Team.) 
 
+(This timeline can be adjusted as needed, at any time, by consensus of the Interop Team.) 
+
 ### Call for Proposal Period
 
 An open call for proposals will begin on a specific date, and run for three weeks.
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -70,7 +70,7 @@ Decisions will be made at a specific meeting for discussion of preliminary elimi
 
 At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
 
-Each member organization chooses 40 number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
+Each member organization chooses 40 proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
 
 The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
 
--- a/2024/planning-process.md
+++ b/2024/planning-process.md
@@ -29,7 +29,7 @@ An open call for proposals will begin on a specific date, and run for three week
         1. [IETF](https://www.ietf.org/) (Standard Track)
         2. [Khronos Group](https://www.khronos.org/)
         3. [TC39](https://tc39.es/) (Stage 2+)
-        4. [W3C](https://www.w3.org/) (Recommendation Track)
+        4. [W3C](https://www.w3.org/) ([Recommendation Track](https://www.w3.org/2023/Process-20230612/#rec-track))
         5. [WHATWG](https://whatwg.org/)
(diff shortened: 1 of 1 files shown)


2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Call for Proposal Period
+
+An open call for proposals will begin on a specific date, and run for three weeks.
+
+1. Anyone can submit a proposal.
+2. Proposals should be for either:
+    1. Focus Area, or
+    2. Investigation Project
+3. Related proposals should be filed separately. For example, all the ideas for “Web Compat” should be filed as a set of separate issues, not one issue. This makes it easier to discuss and prioritize each separately. Bundles of proposals risk being rejected or deprioritized.
```
jgraham, line 76: So, broadly that's what we did the first year for webcompat, and then we switched in 2023. One reason was that sometimes the web compat proposals have been very small; possibly literally just one bug. At that level it seems like a lot of overhead to file multiple issues. On the other hand the process in 2023 wasn't perfect either and there was some confusion about what was included. so I don't object to changing back, but my concern is that people end up filtering out individual proposals because they're "too small" when in fact the intent was to group them all along. This particularly applies if there's a fixed size for the priority list below.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+4. Focus Area proposals must:
+    1. Name a specific web technology.
+        1. This can be a web technology that has not yet been implemented in any or all browser engines.
+        2. Or this can be a web technology that has already been implemented in all three engines, but those implementations have multiple test failures that reveal a lack of compliance to the web standard, causing a lack of interoperability.
```
jgraham, line 80: Are multiple test failures actually a requirement? If there's a compat issue that's encapsulated in a single bug I don't think I should have to write a new test to to reach the arbitary requirement of "multiple failures" (in general I agree with the idea that final focus areas will have multiple test failures).

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+After the Call of Proposals Period has ended, the organizing committee will finalize an official list of all Focus Area proposals.
+
+Any Focus Area proposal that does not meet the minimum publicly-stated selection criteria will be considered for preliminary elimination. For example:
+
+1. The proposal is overly vague or broad.
+2. The proposed technology is already interoperable according to WPT results.
+3. The proposed technology is not on a standards track, or its standard is not sufficiently mature.
+
+Decisions will be made at a specific meeting for discussion of preliminary eliminations. Any member organization can bring up a proposal they believe should be eliminated. Brief debate can be had. If **there is quick consensus**, that proposal will be eliminated. If there is not consensus and a longer debate is needed, the proposal will be not be eliminated and will instead move on to the next round.
```
jgraham, line 113: It's unclear to me what the value of not eliminating if anyone proposes elimination is. Or at least: if someone makes it clear that they will veto a proposal in later rounds it may as well be eliminated early.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+3. Brief debate on each proposal:
+    1. if there is quick consensus, disqualify the proposal
+    2. otherwise, send the proposal to Round 2
+4. Publish the list of proposals pushed to Round 2.
+
+### Elimination Round 2: Reduction
+
+At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
+
+Each member organization chooses `N` number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
```
jgraham, line 126: I don't understand how this will work in detail.

Incoming proposals may be different sizes. Per the above text some proposals may be far too small to form a focus area on their own. So in order to do fixed-N selection I think we'd need to have already grouped the proposals into similar sized chunks at this point. Otherwise it's impossible to compare proposals like "fix one small but important bug" and e.g. "implement an entire new CSS layout mode". 

Also it's unclear how to pick N. We probably need N large enough that we're guaranteed to get enough proposals that Interop 2024 is a reasonable sized project. But given overlaps and possible later vetos, that doesn't seem possible without allowing the process to be iterative. One point of view is "well even if we only end up with five focus areas, we'll know those are the most important ones". But I think that would be a mistake if we end up rejecting things that everyone does in fact end up working on just because Process.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+(If there appear to be quite a few proposals remaining in this round, the group can decide to only ask for each organization to submit their Top `X` number, while rating the rest zero. For example, if 50 proposals remain at this point, it likely makes sense to only submit a Top 25, and to not worry about ranking 26-50 in a specific order.)
+
+After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom.
+
+NOTE: in the past, “Exclude” was a way to express many different kinds of sentiments — including: does not meet the minimum qualifications, or this is low priority to us. We expect “veto” to be used much less than “exclude”, because there are other ways to express those sentiments. Vetoing is specifically for objecting to a particular proposal being included *at all,* no matter how highly ranked other organizations might find it. It’s for expressing objection to a particular web technology, perhaps to something an organization has already expressed an objection to in the web standards process or otherwise has an objection to implementing.
```
jgraham, line 148: I don't think we should try to make this distinction. I think it's totally reasonable to exclude things for any reason, not just because they shouldn't have made it to this point in the process. I strongly disagree with including any language that makes it seems that "exclude" means "we object to this as a web technology".

jgraham (changes requested):

2024/planning-process.md
jensimmons: The word multiple can easily be stricken. I don't think we intended to exclude something that might only have one test failure. More that we were just assuming anything that someone would proposed does have multiple failures.

2024/planning-process.md
jensimmons: We can remove this language from the official process. I believe we initially wrote it to be more of an explanation to the group of member orgs, when first proposing this overall idea. To say, hey, it's likely we won't use this veto in all the ways we used 'exclude' in the past. But of course, yes, any organization is free to use or not use a veto in whatever way they wish.

2024/planning-process.md
jensimmons: N should be "slightly too large". 

If you imagine we expect 35 proposals and 20 "open slots", then figuring out N does seem impossible. And honestly, this whole step could be skipped. Just take all 35 proposals to the next round. 

But if you instead imagine we could get 150 proposals and have only 5 open slots... then this step makes more sense. It's a way to make a huge cut very quickly to get the number of proposals into a reasonable number. Why sweat the details over which is ranked 82nd and which is ranked 97th if we are only accepting 9 new areas? 

We won't know how many open slots there are, but we could guess, for instance, it's 15 at most, as so... let's make N be... 25? Or 40? And cut the other 100+ proposals (if that many come in).

> We probably need N large enough that we're guaranteed to get enough proposals that Interop 2024 is a reasonable sized project. 

Exactly. And it's fine if N is "too high". It's really just a protection against have way too many proposals in the later stage of the process, just in case the success of and attention on Interop 2021-23 brings us exponentially more proposals yet again this year.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Elimination Round 3: Prioritization
+
+The official list of Focus Area proposals is updated to reflect the elimination of proposals not chosen by any organization. Each organization now has 20 days to review the list of remaining proposals and make their Round 3 decisions.
+
+In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise.
+
+Any proposal that gets a veto is immediately eliminated.
+
+Each proposal gets assigned points from each organization, according to their rank, and an ordered list of all remaining proposals is created. Where `X` is number of proposals that are ranked, and `Y` is the ranking a particular proposal gets from a certain organization, then that proposal gets `X-Y`  points from that organization. Once all 6 rankings are in for that proposal, the total points from all six organizations is known. If that proposal is not eliminated by a veto, then it has a certain number of points, and can be listed in order among the other proposals.
```
chrishtr, line 140:
```suggestion
Each proposal gets assigned points from each organization, according to their rank, and an ordered list of all remaining proposals is created. Where `X` is number of proposals that are ranked, and `Y` is the ranking a particular proposal gets from a certain organization, then that proposal gets `X-Y+1`  points from that organization. Once all 6 rankings are in for that proposal, the total points from all six organizations is known. If that proposal is not eliminated by a veto, then it has a certain number of points, and can be listed in order among the other proposals.
```

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+* * *
+
+# Decision point — what to do about Interop 2023?
+
+## There seem to be 3 options:
+
+### 1. Assume carryover.  Retire some.  Add how many ever that leaves space for.
+
+1. Start with the assumption that all Interop 2023 projects will be carried over — except for the ones that are “complete”
```
chrishtr, line 164: I like this option. Sometimes things take longer than we expected, and by default we should try to follow through, even if takes longer than a year.

2024/planning-process.md
chrishtr, line 148:
```suggestion
NOTE: in the past, “Exclude” was a way to express many different kinds of sentiments — including: does not meet the minimum qualifications, or this is low priority to us. We expect “veto” to be used much less than “exclude”, because there are other ways to express those sentiments. Vetoing is specifically for objecting to a particular proposal being included *at all,* no matter how highly ranked other organizations might find it. It’s for expressing objection to a particular web technology, perhaps to something an organization has already expressed an objection to in the web standards process or otherwise has an objection to implementing. A veto should when possible come with a brief explanation to the other Interop Team members to help them understand the reasoning, and further encourage scoring rather than vetoes.
```

chrishtr (commented):
Thanks for the detailed proposal! Most of my comments are nits and phrasing adjustments.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+If there is a particular blocker making it hard to test a certain feature, that could make for a good Investigation Project proposal. And ideas about how the WPT/Interop Project testing infrastructure can be improved for the future can be proposed as Investigation Projects, to be considered as work for the Interop Project to take on in the next year.
+
+### **Requirements for Focus Area proposals**
+
+* This feature has a mature-enough web standard from an established standards development organization such as IETF, Khronos Group, TC39, WHATWG, or the W3C.
+* This feature has high quality tests in WPT or Test262.
+
+### **Guidance for Prioritizing Focus Area proposals**
+
+We should prioritize areas where web developers have had problems using them because of differences between browsers, or because they aren’t implemented in all browsers. There are many sources for such information:
```
boazsender, line 34: Can we say: 

Beyond the above requirements, companies participating in Interop 2024 will consider areas where web developers have had problems using a technology because of differences between browsers, or because they aren’t implemented in all browsers. Participating companies will use the following sources as inputs to each company's internal prioritization process:

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+* Perhaps there are issues filed about such problems in the Chromium, Gecko, or WebKit bug trackers.
```
boazsender, line 36: Can we remove "Perhaps there are..." from each line here.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+* Perhaps there are many complaints about this feature in public discussions on sites like Stack Overflow.
+* Perhaps this feature ranks in high-demand or as a pain-point consistently in polls and surveys of web developers.
+* Perhaps there are documented examples of sites or libraries working around the fact that it is missing or not interoperable.
+
+When assessing the prioritization of existing web technology, it can be helpful to assess how often the technology is being used currently. (Such thinking should also consider important technology might not be in popular use if a lack of interoperability is a painful barrier.)
```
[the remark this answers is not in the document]
boazsender: Can we say:

"Participating companies will also look at how often the technology is being used by evaluating the following sources:"

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+* The feature has high usage via Chrome use counters.
+* The feature has high usage via HTTP Archive.
+* The feature has a high number of mentions on developer resources like Stack Overflow.
+
+There are also guiding values that should be taken into consideration, beyond pure demand:
+
+* This feature has a positive impact on accessibility.
```
boazsender, line 49: s/This/The for consistency

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+* This feature has high quality tests in WPT or Test262.
+
+### **Guidance for Prioritizing Focus Area proposals**
+
+We should prioritize areas where web developers have had problems using them because of differences between browsers, or because they aren’t implemented in all browsers. There are many sources for such information:
+
+* Perhaps there are issues filed about such problems in the Chromium, Gecko, or WebKit bug trackers.
+* Perhaps there are many complaints about this feature in public discussions on sites like Stack Overflow.
+* Perhaps this feature ranks in high-demand or as a pain-point consistently in polls and surveys of web developers.
+* Perhaps there are documented examples of sites or libraries working around the fact that it is missing or not interoperable.
```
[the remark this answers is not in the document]
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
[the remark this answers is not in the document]
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
```diff
@@ -0,0 +1,187 @@
...
+
+When assessing the prioritization of existing web technology, it can be helpful to assess how often the technology is being used currently. (Such thinking should also consider important technology might not be in popular use if a lack of interoperability is a painful barrier.)
+
+* The feature has high usage via Chrome use counters.
+* The feature has high usage via HTTP Archive.
+* The feature has a high number of mentions on developer resources like Stack Overflow.
+
+There are also guiding values that should be taken into consideration, beyond pure demand:
```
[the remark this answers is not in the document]
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
gsnedders: I've attempted to address this in #394.

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+1. Setup *Disqualification meeting*
+2. Each organization can propose items for disqualification.
+3. Brief debate on each proposal:
+    1. if there is quick consensus, disqualify the proposal
+    2. otherwise, send the proposal to Round 2
+4. Publish the list of proposals pushed to Round 2.
+
+### Elimination Round 2: Reduction
+
+At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
```
jgraham, line 71: At this point there's a clear list of proposals. But if focus areas can be groups of proposals, there isn't a clear list of focus areas.

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+    3. That determines how many “open slots” we have.
+        1. 26 - the number retired = the number carried over.
+        2. If we are aiming for 20 areas in 2024, and 18 are carried over (retiring 6)... then we have space to add 2 new areas.
+
+### 2. Automatically ‘reapply’ all Focus Areas
+
+1. Carryover nothing.
+2. Take all 26 Focus Areas from Interop 2023, and put them into the selection process.
+
+### 3. Make every Focus Area ‘reapply’
```
jgraham, line 125: I prefer this option.

I agree that previous focus areas should automatically be given weight by orgs when considering priorities. But I also think there should be a real assessment of the value of completing existing focus areas vs adding new ones. In some cases this will be obvious (e.g. if a feature is still missing from some implementations), and in some cases it won't be (e.g. if there's a few percent left, and work is needed to understand if the remaining work carries its weight vs other options). In some cases there might be a third way e.g. grouping the remaining work from several previous focus areas into a new focus area with just the remaining failures (e.g. one can imagine flexbox + grid + subgrid failures becoming a single "Layout modes" focus area).

Re-proposing is more work, but I think that work has enough value to make it worthwhile.

jgraham (changes requested):

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+
+If `N` is increased, then more proposals will make it into the next round. The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
+
+### Elimination Round 3: Prioritization
+
+The official list of Focus Area proposals is updated to reflect the elimination of proposals not chosen by any organization. Each organization now has 20 days to review the list of remaining proposals and make their Round 3 decisions.
+
+In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise.
```
chrishtr, line 83:
```suggestion
In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise, with a brief rationale for the benefit of the Interop team.
```

jgraham: I don't think the process can require a reason for exercising a veto. Of course I also hope in most cases people will be able to articulate reasoning, but nevertheless.

2024/planning-process.md
nairnandu: +1 to jgraham's comment above about selecting 'N' being particularly tricky. Suggest each organization pick a reasonable number of proposals that are well-justified and that they can support.

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Elimination Round 1 : Disqualification
+
+After the Call of Proposals Period has ended, the organizing committee will finalize an official list of all Focus Area proposals.
+
+Any Focus Area proposal that does not meet the minimum publicly-stated selection criteria will be considered for preliminary elimination. For example:
+
+1. The proposal is overly vague or broad.
+2. The proposed technology is already interoperable according to WPT results.
+3. The proposed technology is not on a standards track, or its standard is not sufficiently mature.
```
bkardell, line 58: Is the technical inability to currently test something a potential reason for disqualification?

2024/planning-process.md
bkardell: +1

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+
+During the same period organizations are making Round 3 decisions, the organizing committee should evaluate the status of Interop 2023, and decide which Focus Areas will be retired, and which will be carried over.
+
+\[XXX]
+
+* * *
+
+# Decision point — what to do about Interop 2023?
+
+## There seem to be 3 options:
```
jgraham, line 107: OK, so based on the discussion, here's an alternative option:

1. Do an (early) post-mortem for all Interop 2023 proposals. File an issue on GitHub for each one, detailing which tests are not yet interoperable (if any).
1. Encourage participants to comment on the focus area. For example they might want to include:
    * Information about whether significant changes are planned/expected in the near future in their implementations.
    * Additional context they can share around remaining failures (e.g. reasons that they turned out to be lower priority or harder to fix).
    * Any additional coverage that's missing from the focus area and should be included if the focus area is carried forward.
1. Given this information, we compile a list of candidate focus areas to carry forward. This would include any where we believe there is still meaningful Interop work to be completed, either because we can see the remaining failures are meaningful (i.e. it's not a collection of unlikely edge cases), or because we believe that the tests/scope of the focus area should be increased.
1. At proposal selection time, we work from both the list of potential carry-over and the list of new proposals in order to fill the project out to the desired size. There is no commitment to take everything on the potential carry-over list, but there is an assumption that items will be carried over if they are legitimately seen as ongoing priority areas for improving Interop.

This has the following benefits over any of the existing proposals:
* It encourages us to make an evidence-based decision on the impact of remaining failures in existing focus areas, rather than assuming some fixed numerical cut off, or making an ad-hoc decision based on gut feelings.
* It explicitly allows for the possibility of these focus areas being expanded with new tests that may cover additional scenarios which weren't included in 2023.
* It allows us to provide explicit priority to carryover proposals, without requiring that any carryover always outranks new proposals, as in the current option 1.

Of course this would still be subject to the usual consensus-based process for final decisions.

2024/planning-process.md
chrishtr: > I think it seems pretty easy to present, have the group accept that and move on

Nevertheless this will take time, and it's hard enough to make decisions already. The purpose of rounds 1 and 2 is about efficiency of easy decisions, so we should stick to easy decisions and not ones where there is any debate at all.

2024/planning-process.md
gsnedders: I think we need to get agreement on the more fundamental disagreement first: are we requiring consensus to carryover *or* are we require consensus to retire? Because what we should do process-wise likely depends on that decision.

2024/planning-process.md
gsnedders: Yeah, as @chrishtr says—it's about making non-controversial decisions quickly.

It's also intended that the first round doesn't have too many subjective decisions—a spec either does or does not meet the criteria we've set (and I hope we don't get into disagreements here!), tests already show interoperability or they don't (we can of course get into a debate about whether the final 5% matters!), though "overly vague or broad" is much more so subjective.

@jgraham: was your original question intended to be "why not let people veto up front, for any reason"? if so—the current kinda short list of reasons for disqualification are there because they're largely what we practically did in prior years, and because things that are largely objective are likely easier to get consensus on.

2024/planning-process.md
gsnedders: One of the big questions that came up in the meetings in recent weeks was the question of when to do the grouping; it might be worth revisiting this after deciding when to do the grouping?

2024/planning-process.md
bkardell: Is it possible that vagueness is causing us to be talking about completely different things?  I think yes let's ask people to 'veto' here, and not at the end.  That is what I am saying - I might be completely alone in this... But honestly, I also kind of dislike that we're even using words like "vote" and "veto" at all because they conjure up something that I don't honestly think is really happening here.  I understand why, it's a kind of vote - but really what we're doing is more like using some joint worksheets to do complicated sorting and planning and negotiate (with what little power any of us actually has to do so) some plausible subset of things we all agree we could work on improving interoperability of.

Let's say two concrete examples:

* Somebody submits that we should work on interoperability of the `is=""` attribute.  It's in the WHATWG HTML spec, which I think has always had the same definition "two in favor, none opposed".  Except Apple _is_ opposed, and has a public position many times stating as much, including a recent WebKit standards position.  So... Is it disqualified in that case? Or does this require a "veto"?  If the latter, we can say it now, why drag it out.  It's just a particularly easy example, but it's not unique: People are going to submit things that have negative standards positions - is this the venue to debate it? I kind of think not.

* Somebody submits we should work on 'all the color stuff' a few years ago and it's pointed out that this requires much low level rework in order to actually bring the value, so it's not likely something that would actually improve this year and there are lots of _other_ things where we think we could all agree and get stuff measurably done and shipped --- but hey we are actually working on all that low level stuff and totally support it!  That isn't currently disqualified, and it's certainly not a "veto" - but it's a known known and it seems to me the group could choose to say "that seems totally reasonable, let's just exclude it" and doing that as early as possible just seems good... To me.

jgraham: Let's say someone proposes "Geolocation Sensor" for Interop 2024. It has a [spec](https://w3c.github.io/geolocation-sensor/) on a standards track at W3C, and [tests](https://github.com/web-platform-tests/wpt/tree/master/geolocation-sensor). So per the criteria given we can't eliminate the proposal at this stage. But it also has [multiple](https://github.com/WebKit/standards-positions/issues/32) [negative](https://github.com/mozilla/standards-positions/issues/36) standards positions, and no evidence of implementations going ahead despite those standards positions.

It seems like a pretty safe bet to me that proposal is not going to reach consensus.

But per the process here, we have to take it forward in case we want to have a longer debate at some later stage. But it's not clear to me what form that debate could possibly take, and indeed debating proposals as a group isn't part of any of the later stages of the proposed process. So assuming at least one org is in favour in round 2, the proposal has to go all the way through to round 3, even though we had the information in round 1 to know it wasn't going to reach consensus (because a participant could have said so).

Obviously this isn't the most serious concern; eventually we'll incorporate the information that the proposal isn't going to reach consensus, and it won't get selected. But it is an inefficiency in a process that claims to be optimising for making the best use of everyone's time.

2024/planning-process.md
jgraham: I don't think it makes sense to talk about consensus except in the positive sense i.e. complete agreement rather than complete disagreement.

If we don't have consensus that something actively should be in Interop 202x then we don't have consensus to publish that iteration of the metric.

2024/planning-process.md
jgraham: How to do any grouping, and how to estimate/balance the overall size of the focus areas, seem like the most fundamental remaining issues. Previously we did that at the last minute by "feel", but that doesn't work well with the process set out here due to the ordering step.

 Grouping before ordering doesn't work if that happens before the proposals have got detailed input from engineering teams, because we might group things that have very different levels of complexity or priority, and then be forced to assess the group as a whole on the basis of its worst elements.
 
 But trying to order things of very different size and complexity also doesn't seem to work well (how does one tradeoff "fix a single test that represents a serious webcompat issue but is probably a few days of work and isn't going to be a focus area on its own" against "implement an entire new platform feature that could take multiple months of work"?).

foolip: The suggestion I like the best so far is this one from @jgraham:

> ask people to pick a number of proposals that they feel reflects a reasonable size for Interop 2024

I think there's general agreement that we should not increase the number of focus areas, so at the very very most this could be 26.

This does require each organization to make some qualified guesses about what will make sense to group after round 3. In the case of web compat bugs this is easy. For some other proposals there might be disagreement, but the only effect of this is that a few more proposals move on to round 3.

Corner case consideration: Elsewhere I suggested we should also allow for vetos in round 2. If we do this, then there's no guarantee that we have N proposals after round 2. I think this is very unlikely, but if it does happen we could do another round of round 2 for the remaining slots. Presumably there would be no vetos that time.

2024/planning-process.md
foolip: Indeed we can't require a reason, how about we make it clearly optional but encouraged:

```suggestion
In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise, if possible with a brief rationale for the benefit of the Interop team.
```

2024/planning-process.md
foolip: I can see the argument for both options:
- Requiring consensus to carryover makes sense since any objection means someone organization isn't happy with the metric, and over time this can threaten the viability of the project.
- Requiring consensus to retire makes sense since we should finish what we already agreed to, and not relitigate.

But I'm most convinced by the first argument. I don't think there's a significant risk of us dropping things that are still very important with this setup.

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Timeline — 10 weeks total
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 5)
+2. Continued discussion and refinement of proposals — one week (Oct 5 – Oct 12)
+3. Elimination Round 1: Disqualification — one week (Oct 12 – Oct 19)
+4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
+5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
+
+
```
foolip, line 13:
```suggestion
Interop 2024 and the set of accepted proposals will be announced in January 2024.
```

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
+# Proposal for Interop 2024 Process
+
+## Process
+
+### Timeline — 10 weeks total
```
foolip, line 5:
```suggestion
### Timeline
```

foolip (commented):
Adding a date based on in-person discussion in https://github.com/web-platform-tests/interop/issues/405.

foolip:
Let me try to enumerate the remaining issues that we need to resolve here:

- Process for carrying over focus areas from 2023 into 2024. @jgraham suggested filing an issue for each focus area and discussing there if the remaining test failures are important or not. I think we should do that, and think it'll be easy-ish to agree about the carry over afterwards.
- In round 2, remove or define the limit _N_ on the number of proposals each organization can pick. I suggest we remove it and go with "ask people to pick a number of proposals that they feel reflects a reasonable size for Interop 2024" as suggested by @jgraham. Also allow each org to group proposals as they think would make sense in this round.
- Define what's member-confidential, leaning on https://github.com/web-platform-tests/interop/pull/102. @gsnedders @jgraham can either of you suggest the wording you'd prefer?

Am I missing anything?

gsnedders:
> Am I missing anything?

* When in the process do we do the grouping into focus areas?

foolip:
> When in the process do we do the grouping into focus areas?

I'd suggest that each organization say in round 2 which proposals they're thinking will be grouped, of the proposals they're supporting. Then after round 3 we'll have to come to a joint team decision about it similar to in past years.

foolip:
Additionally, pointed out by @jensimmons:

- Ensure we're happy with the wording used for positions

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+3. Brief debate on each proposal:
+    1. if there is quick consensus, disqualify the proposal
+    2. otherwise, send the proposal to Round 2
+4. Publish the list of proposals pushed to Round 2.
+
+### Elimination Round 2: Reduction
+
+At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
+
+Each member organization chooses 40 number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
```
dandclark, line 73:
```suggestion
Each member organization chooses 40 proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+For example, if there are 20 remaining proposals, each organization will choose a number 1, number 2, down to number 20. Let’s say a proposal for flying cars gets the following ranks: 1, 5, 7, 10, 14, 20. That will give it the following points: 20, 15, 13, 10, 6, 1 = 65 points total.
+
+(If there appear to be quite a few proposals remaining in this round, the group can decide to only ask for each organization to submit their Top `X` number, while rating the rest zero. For example, if 50 proposals remain at this point, it likely makes sense to only submit a Top 25, and to not worry about ranking 26-50 in a specific order.)
+
+After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom.
```
foolip, line 91: Suggesting an additional bit after a discussion in the team meeting today:

```suggestion
After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom. The team will decide by consensus which of the proposals will be included in Interop 2024 and how they will be grouped into focus areas and investigation efforts.
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+3. Brief debate on each proposal:
+    1. if there is quick consensus, disqualify the proposal
+    2. otherwise, send the proposal to Round 2
+4. Publish the list of proposals pushed to Round 2.
+
+### Elimination Round 2: Reduction
+
+At this point, there is a clear list of Focus Area proposals under consideration. Each organization has two weeks to review the list of proposals and make Round 2 decisions.
+
+Each member organization chooses 40 proposals they would like to see move on to Round 3; these selections are confidential to the team. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
```
dandclark, line 73: Regarding the confidentiality wording added in [7093980](https://github.com/web-platform-tests/interop/pull/390/commits/70939802be67b4354a0436271cdf13b923bdcc1d), there's a proposal I'd floated a couple meetings ago that I'd like to revisit in the interest of nudging the process towards openness where possible:

> Once the announcement of selected proposals has been made, members may choose to share publicly the anonymized, cumulative results of each round. Specifically, that is: the set of proposals that were disqualified in round 1, the number of "yeses" that each proposal received in round 2, and the number of points that each proposal received in round 3.

This would maintain the confidentiality of individual member organizations' positions, which I hope addresses concerns raised about positions being taken out of context.

There was also a question raised about whether this is worth the trouble for the Interop team to report. I don't think there's much work involved since we will need to compute these totals anyway as part of the process, but nevertheless I think that concern is addressed by making the sharing of this info an option for members rather than an obligation of the Interop team.

Thoughts?

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+After the Call of Proposals Period has ended, the Interop Team will finalize an official list of all Focus Area proposals.
+
+Any Focus Area proposal that does not meet the minimum publicly-stated selection criteria will be considered for preliminary elimination. For example:
+
+1. The proposal is overly vague or broad.
+2. The proposed technology is already interoperable according to WPT results.
+3. The proposed technology is not on a standards track, or its standard is not sufficiently mature.
+
+Decisions will be made at a specific meeting for discussion of preliminary eliminations. The minutes of this meeting are confidential to the team. Any member organization can bring up a proposal they believe should be eliminated. Brief debate can be had. If **there is quick consensus**, that proposal will be eliminated. If there is not consensus and a longer debate is needed, the proposal will be not be eliminated and will instead move on to the next round.
```
jgraham, line 60: I'm not convinced that the phrasing that it's specifically the "minutes of the meeting" which are confidential captures the intent. AIUI the intent is that vendor positions are confidential throughout, irrespective of how they're shared (e.g. if the data is captured in a spreadsheet outside of the context of a specific meeting).

Although the document does also say that those minutes are shared in a specific meeting, in practice I think it's likely that we'll want some kind of async method to collate and share information.

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
+
+### Elimination Round 3: Prioritization
+
+The official list of Focus Area proposals is updated to reflect the elimination of proposals not chosen by any organization. Each organization now has 20 days to review the list of remaining proposals and make their Round 3 decisions.
+
+In this round, each organization ranks the proposals in order of importance from most to least. And submits any vetos they may wish to exercise. The entire contents of this round is confidential to the team.
+
+Any proposal that gets a veto is immediately eliminated.
```
jensimmons, line 83:
```suggestion
Any proposal that has an objection is immediately eliminated.
```

foolip:
The only remaining work here for me to give an approving review is to remove the final section about carryover, or replace it with a short description of what we've concluded in meetings. I think something this would suffice:

> All focus areas in Interop 2023 will be evaluated for carryover into Interop 2024. Only those focus areas which have consensus to continue will be part of Interop 2024. As part of this evaluation, the scope may also be changed, and focus areas can be merged or split when there is consensus to do so.

I don't think we need to say anything about the logistics (filing issues).

@gsnedders WDYT?

jensimmons:
I've moved this draft to GDocs so that it's easier to edit:
https://docs.google.com/document/d/138F0bzFwywci9p5q7HMP4Rbi9E0iBjz4s0ZC6IE9324/edit?usp=sharing

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
+# Proposal for Interop 2024 Process
+
+## Process
+
+### Timeline
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 5)
```
jensimmons, line 7:
```suggestion
1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+2. Continued discussion and refinement of proposals — one week (Oct 5 – Oct 12)
```
jensimmons, line 8:
```suggestion
2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
+# Proposal for Interop 2024 Process
+
+## Process
+
+### Timeline
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
+2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
+3. Elimination Round 1: Disqualification — one week (Oct 12 – Oct 19)
```
jensimmons, line 9:
```suggestion
3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
+# Proposal for Interop 2024 Process
+
+## Process
+
+### Timeline
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
+2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
+3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
+4. Elimination Round 2: Reduction — two weeks (Oct 19 – Nov 2)
```
jensimmons, line 10:
```suggestion
4. Elimination Round 2: Reduction — two weeks (Oct 23 – Nov 3)
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Process
+
+### Timeline
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
+2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
+3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
+4. Elimination Round 2: Reduction — two weeks (Oct 23 – Nov 3)
+5. Elimination Round 3: Prioritization — three weeks (Nov 2 – Nov 30, exc. North America Thanksgiving)
```
jensimmons, line 11:
```suggestion
5. Elimination Round 3: Prioritization — three weeks (Nov 6 – Nov 22)
```

2024/planning-process.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Timeline
+
+1. Call for Proposal Period — three weeks (Sep 14 – Oct 7)
+2. Continued discussion and refinement of proposals — one week (Oct 8 – Oct 14)
+3. Elimination Round 1: Disqualification — one week (Oct 16 – Oct 20)
+4. Elimination Round 2: Reduction — two weeks (Oct 23 – Nov 3)
+5. Elimination Round 3: Prioritization — three weeks (Nov 6 – Nov 22)
+
+Interop 2024 and the set of accepted proposals will be announced in January 2024.
```
jensimmons, line 13:
```suggestion
Interop 2024 and the set of accepted proposals will be announced in January 2024.

(This timeline can be adjusted as needed, at any time, by consensus of the Interop Team.) 
```

2024/planning-process.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Call for Proposal Period
+
+An open call for proposals will begin on a specific date, and run for three weeks.
+
+1. Anyone can submit a proposal.
+2. Proposals should be for either:
+    1. Focus Area, or
+    2. Investigation Project
+3. Related proposals should be filed separately. For example, all the ideas for “Web Compat” should be filed as a set of separate issues, not one issue. This makes it easier to discuss and prioritize each separately. Bundles of proposals risk being rejected or deprioritized.
```
jensimmons, line 27:
```suggestion
3. Related proposals should be filed separately. For example, a group of specific features to address under “Typography” should be filed as separate issues, not one issue. This makes it easier to discuss and prioritize each separately. Bundles of proposals risk being rejected or deprioritized.
```

2024/planning-process.md
jensimmons: Changing to be a different example, since, right, Web Compat is likely an exception to this guideline.

2024/planning-process.md
jensimmons: Drafting up this instead (in the GDoc):
> The positions of individual member organizations are confidential to the team.

2024/planning-process.md
```diff
@@ -0,0 +1,136 @@
...
+3. The proposed technology is not on a standards track, or its standard is not sufficiently mature.
+
+Decisions will be made at a specific meeting for discussion of preliminary eliminations. The minutes of this meeting are confidential to the team. Any member organization can bring up a proposal they believe should be eliminated. Brief debate can be had. If **there is quick consensus**, that proposal will be eliminated. If there is not consensus and a longer debate is needed, the proposal will be not be eliminated and will instead move on to the next round.
+
+1. Setup *Disqualification meeting*
+2. Each organization can propose items for disqualification.
+3. Brief debate on each proposal:
+    1. if there is quick consensus, disqualify the proposal
+    2. otherwise, send the proposal to Round 2
+4. Publish the list of proposals pushed to Round 2.
```
jensimmons, line 71: Wait what? In public? We should debate this. Please see GDoc for more. https://docs.google.com/document/d/138F0bzFwywci9p5q7HMP4Rbi9E0iBjz4s0ZC6IE9324/edit#heading=h.jeelvxhy0wl3

2024/planning-process.md
jensimmons: Yes. Done in the GDoc.

2024/planning-process.md
```diff
@@ -0,0 +1,134 @@
...
+
+Each member organization chooses `N` number of proposals they would like to see move on to Round 3. There is no ranking or scoring at this point, a simple “yes, we want this proposal to stay in consideration”. Once all the “yeses” are gathered, any proposal that does not have the support of at least one organization is eliminated.
+
+For example, let’s imagine there are 112 focus area proposals submitted. Let’s set `N` to 10 — so each of the six organizations gets to choose 10 proposals. There will likely be significant overlap, but not complete. So imagine 32 proposals get chosen by at least one organization. (It will be no less than 10, and no more than 60.) This means the remaining 80 proposals are eliminated.
+
+If `N` is increased, then more proposals will make it into the next round. The goal is to find a balance so all of the proposals that are most likely to be considered high priority will make it into the next round. While eliminating as many as possible, to reduce the workload of the Interop 2024 organizations. There’s no need to extensively evaluate proposals of lower priority.
+
+### Elimination Round 3: Prioritization
```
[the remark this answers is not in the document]
jensimmons: Do you still have this concern, James? If so, any suggestions about how to address it? (I recommend reading the current language in the GDoc, and commenting there.)

2024/planning-process.md
```diff
@@ -0,0 +1,136 @@
...
+
+After Round 3, every remaining proposal can be listed in order, with the group’s highest priority at the top, and the lowest at the bottom. The team will decide by consensus which of the proposals will be included in Interop 2024 and how they will be grouped into focus areas and investigation efforts.
+
+NOTE: in the past, “Exclude” was a way to express many different kinds of sentiments — including: does not meet the minimum qualifications, or this is low priority to us. We expect “veto” to be used much less than “exclude”, because there are other ways to express those sentiments. Vetoing is specifically for objecting to a particular proposal being included *at all,* no matter how highly ranked other organizations might find it. It’s for expressing objection to a particular web technology, perhaps to something an organization has already expressed an objection to in the web standards process or otherwise has an objection to implementing.
```
jgraham, line 97:
```suggestion
```

I still think this note, whilst non-normative, is nevertheless misleading and that it remains acceptable for participants to veto any proposal for any (public or private) reason. Although I agree with the sentiment that direct vetos are less likely in the proposed process, I don't think the Process should try to hint at the existence of legitimate and illegitimate categories of reasons to veto, particularly given that we don't have experience of how things will play out in practice, and so don't know what will actually be required to end up with a final result that everyone can approve.

(If this PR is merged, as per the previous meeting notes, I will reopen this as a new PR).

2024/planning-process.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Interop 2023 Evaluation
+
+During the same period organizations are making Round 3 decisions, the Interop Team should evaluate the status of Interop 2023, and decide which Focus Areas will be retired, and which will be carried over.
+
+\[XXX]
```
jgraham, line 103: This appears to still be placeholder text?

2024/planning-process.md
jensimmons: This has already been removed. Have you checked the Google Doc draft? That's where the current draft is. This PR is now irrelevant.

2024/planning-process.md
jensimmons: This has been addressed in the current draft, in GDocs.

jgraham:
Ah, apologies, the meeting notes suggested that the PR was going to be merged, so I assumed this was the canonical version again.

jensimmons:
I'm going to close this PR, because we've created a new version of the proposal, which is now at 
https://github.com/web-platform-tests/interop/pull/591

Issue #393, opened 2023-08-17
Agenda, Aug 17th 2023


chrishtr:
1. Review Apple's proposal for the 2024 test selection process:
https://github.com/web-platform-tests/interop/pull/390

2. Review James's proposal for the proposal submission template:
https://github.com/web-platform-tests/interop/pull/385

3. Charter approval
https://github.com/web-platform-tests/interop/pull/102

4. Test change proposals
https://github.com/web-platform-tests/interop/issues?q=is%3Aissue+is%3Aopen+label%3Atest-change-proposal

chrishtr:
Notes:

```

PR #394, opened 2023-08-18
Add information about the Interop Project to the README

gsnedders:
Split from #390, based on discussion in #393.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,9 @@ Welcome to the Interop Project, an ongoing effort to make the web more interoper
 
 Please see the [Interop 2023](./2023/README.md) for the current iteration of the project!
 
+## Interop 2024
+
+Interop 2024 will open for proposals between Sep 14 and Oct 5, and the Interop Team will evaluate proposals in the following months. Interop 2024 and the set of accepted proposals will be announced in January 2024. Please see the [Interop 2024 Process](./2024/planning-process.md) for details.
 
 ## The Purpose of the Interop Project
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # The Interop Project
 
-Welcome to the Interop project, an ongoing effort to make the web more interoperable in key areas, prioritized by web developer and user needs. This is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+Welcome to the Interop Project, an ongoing effort to make the web more interoperable in key areas, prioritized by web developer and user needs. This is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform, and worked on by the [Interop Team](./supporters.md).
 
 Please see the [Interop 2023](./2023/README.md) for the current iteration of the project!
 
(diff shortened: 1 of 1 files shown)


chrishtr (approved):
This is great, thanks!

README.md
```diff
@@ -1,5 +1,57 @@
 # The Interop Project
 
-Welcome to the Interop project, an ongoing effort to make the web more interoperable in key areas, prioritized by web developer and user needs. This is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform.
+Welcome to the Interop Project, an ongoing effort to make the web more interoperable in key areas, prioritized by web developer and user needs. This is part of [the web-platform-tests project](https://github.com/web-platform-tests/wpt), the main test suite for the web platform, and worked on by the [Interop Team](./supporters.md).
```
jgraham, line 3: I would suggest putting "users" before "web developers" everywhere. Ultimately helping developers is only important insofar as it ultimately benefits users. Generally their needs are aligned (although there may be cases e.g. when features have privacy concerns, where they are not), but I don't think we should imply that developers' needs outweigh those of end users.

README.md
```diff
@@ -1,5 +1,57 @@
...
+
+## The Purpose of the Interop Project
+
+*Improve interoperability significantly for the benefit of web developers and users.*
+
+The goal of the Interop Project is to improve the web by making it easier to make websites and web apps that work in every browser or browser engine at the same time.
+
+This is done by increasing the amount of “interoperability” between browsers — when each browser engine has implemented the same technology the exact same way, as bug-free as possible.
+
+Today’s browsers have made a commitment to implement web technology according to a shared web standard, created in organizations such as the W3C or WHATWG, where technologies such as CSS and HTML are officially defined.
```
jgraham, line 16: Link to the external organisations.

README.md
```diff
@@ -1,5 +1,57 @@
...
+
+There is a seemingly infinite amount of work that browser engineering teams could be focused on. The Interop Project provides incentives to focus on the specific and practical work that will help web developers most in the coming year.
```
jgraham, line 18: I don't think we should frame this as "help web developers most", but maybe something more generic like "have the most positive impact on the web platform"

README.md
```diff
@@ -1,5 +1,57 @@
...
+
+## Scope
+
+The Interop Project is a collaboration between organizations that implement web technology in browser engines. It’s a shared agreement to prioritize work on certain web technologies to improve interoperability. And to keep track of that work by using automated tests to score how much progress each participating browser has made reaching the shared goals.
```
jgraham, line 22: It's not an agreement to prioritise work. It's an agreement to create a public metric. Orgs are of course free to prioritise whatever work they like.

README.md
```diff
@@ -1,5 +1,57 @@
...
+### **Requirements for Focus Area proposals**
+
+* This feature has a mature-enough web standard from an established standards development organization such as IETF, Khronos Group, TC39, WHATWG, or the W3C.
+* This feature has high quality tests in WPT or Test262.
+
+### **Guidance for Prioritizing Focus Area proposals**
+
+Beyond the above requirements, in line with the above goal, areas where web developers have had problems using them should be prioritized, whether that is because of differences between browsers, or because they aren’t implemented in all browsers. There are many potential sources for information about such areas, including but not limited to:
+
+* Issues filed about such problems in the Chromium, Gecko, or WebKit bug trackers.
```
jgraham, line 41: It seems like compat problems observed on real sites should also be mentioned (could be with a link to webcompat.com).

jgraham (changes requested):

README.md
```diff
@@ -1,5 +1,57 @@
...
+* Ranking as a high-demand feature or as a pain-point consistently in polls and surveys of web developers.
+* Documented examples of sites or libraries working around the fact that a feature is missing or not interoperable.
+
+When assessing the prioritization of existing web technology, it can be helpful to assess how often the technology is being used currently. However, one should also consider if important technology might not be in popular use due to lack of interoperability. Potential sources of data include but are not limited to:
+
+* Browser use counters.
+* Usage recorded via HTTP Archive crawls.
+* Number of mentions on developer resources like Stack Overflow.
+
+There are also guiding values that should be taken into consideration, beyond pure demand:
```
boazsender, line 52: I think we should remove ", beyond pure demand" because it makes it sound like these priorities are less important than demand.

chrishtr:
Still LGTM with the comments from James and Boaz addressed. I think both of them made some good suggestions.

README.md
foolip:
```suggestion
There is a seemingly infinite amount of work that browser engineering teams could be focused on. The Interop Project provides incentives to focus on the specific and practical work that will have the most positive impact on the web platform in the coming year.
```

foolip (approved):

jgraham (approved):

PR #415, opened 2023-09-14
Inline Interop 2024 timeline into proposal template

foolip:


foolip (commented):
Reviewed live in meeting

Issue #524, opened 2023-10-04
Agenda for Oct 5th, 2023


nairnandu:
Here is the proposed agenda for the meeting on Oct 5th, 2023

- Follow-up on next steps from https://github.com/web-platform-tests/interop/issues/497, for PR https://github.com/web-platform-tests/interop/pull/390
- Requests for changes to wpt.fyi need to be raised by Nov 2nd, 2023 
- Update timelines on #384 
- Closing the call for proposals
- Interop 2023 [carryover discussion](https://github.com/web-platform-tests/interop/issues?q=is%3Aissue+is%3Aopen+label%3Acarryover-evaluation)

gsnedders:
From last week's agenda:

> Fixing the wording around positions
> * AI: gsnedders will update PR [Add a proposal for the Interop 2024 process #390](https://github.com/web-platform-tests/interop/pull/390) to reflect the wording around positions

It is unclear to me what this action item is meant to encompass, because we don't seem to have minuted which bit of wording is problematic nor how to fix it.

nairnandu:
This was a carry-forward AI from the [Sept 21 meeting](https://github.com/web-platform-tests/interop/issues/453#issue-1905293132). There was a suggestion to change the text on vetos/positions. We did not agree on what it should be, in the meeting.

LeaVerou:
> * Update timelines on [Interop 2024 timeline #384](https://github.com/web-platform-tests/interop/issues/384)

Thanks @foolip! For context, this was about [this comment of mine](https://github.com/web-platform-tests/interop/issues/384#issuecomment-1725932244):

> Hey folks, I just realized that the proposals deadline is Oct 5th, and that the dates in this issue are stale. It would be nice to update them, so that other people don't get confused like I did: we based [our roadmap](https://github.com/Devographics/surveys/discussions/112) on the Oct 13th deadline mentioned here, but it seems to have been changed to [Oct 5th](https://web.dev/interop-2024-proposals/) at some point. The earliest mention of Oct 5th I can find in this repo is [this commit](https://github.com/web-platform-tests/interop/commit/06ebc1b8acc678284f6293ac466011341cca6ad9) and [this one](https://github.com/web-platform-tests/interop/commit/4eac805cb1f99218b2d11bf75541e8b16997ecb1), both ~2 weeks ago.

Not sure if my vote counts, but +1 for following the original schedule!

foolip:
[...]

PR #552, opened 2023-10-05
Extend the proposal window to Oct 7

foolip:


foolip:
As discussed in https://github.com/web-platform-tests/interop/issues/524

foolip:
I missed a place and just pushed https://github.com/web-platform-tests/interop/commit/3528a4485e8e796be0839651f6492eec73a6866c.

Issue #589, opened 2023-10-23
Clarifying spec requirements for Interop 2025


foolip:
The [Interop 2024 proposal template](https://github.com/web-platform-tests/interop/blob/3528a4485e8e796be0839651f6492eec73a6866c/.github/ISSUE_TEMPLATE/focus-area-proposal.yml) said this about the specification:

> The feature must defined by a standards-track specification from one of the following organizations:
>
> - [IETF](https://www.ietf.org/) (Standard Track)
> - [Khronos Group](https://www.khronos.org/)
> - [TC39](https://tc39.es/) (Stage 2+)
> - [W3C](https://www.w3.org/) (Recommendation Track)
> - [WHATWG](https://whatwg.org/)

This issue tracks changes or refinements we want to make. So far:

- Should we accept ISO specs?
- What's the bar for Wasm features?

foolip:
For Wasm, I would suggest [Phase 2+](https://github.com/WebAssembly/proposals#phase-2---proposed-spec-text-available-cg--wg) as the bar.

jensimmons:
Let's add:
- [Alliance for Open Media](https://aomedia.org/)
- [ISO](https://www.iso.org/)

In fact, I'd recommend adding them to the current process proposal. And adding an "etc" to the end. I don't believe we are intending to gate proposals by saying "oh that web standards org doesn't count, only these ones do". It seems our intention is to say "this must be standardized in an industry web standards org". And it's ok if we add some as we go, as we think of technology we want to include. The availability of tests will be more of a blocker than which standards org something is part of. If we someday want to exclude a particular standards org, then we can do so later. 

How about changing the language to this:

> The feature must defined by a standards-track specification from one of the following organizations: 
> - [IETF](https://www.ietf.org/) (Standard Track)
> - [Khronos Group](https://www.khronos.org/)
> - [TC39](https://tc39.es/) (Stage 2+)
> - [W3C](https://www.w3.org/) (Recommendation Track)
> - [WHATWG](https://whatwg.org/)
> - [WASM](https://webassembly.org/specs/) (Phase 2+)
> - [Alliance for Open Media](https://aomedia.org/)
> - [ISO](https://www.iso.org/)
> - or other similar industry web standards organizations

jensimmons:
I created a suggested amendment to the proposal PR at https://github.com/web-platform-tests/interop/pull/591

jgraham:
[...]

PR #591, opened 2023-10-24
Latest proposal for the new Interop process

jensimmons:
The documentation of the new Interop proposal selection process, which we are using for 2024.


2024/selection-process
```diff
@@ -0,0 +1,139 @@
...
+4. Focus Area proposals must: 
+    A. Name a specific web technology. 
+        i. This can be a web technology that has not yet been implemented in any or all browser engines.
+        ii. Or this can be a web technology that has already been implemented in all browser engines, but those implementations have test failures that reveal a lack of compliance to the web standard, causing a lack of interoperability.
+    B. The feature must be defined in a sufficiently mature standards-track web specification from one the following organizations. Please provide a link to the web standard. 
+        i. [IETF](https://www.ietf.org/) (Standard Track)
+        ii. [Khronos Group](https://www.khronos.org/)
+        iii. [TC39](https://tc39.es/) (Stage 2+)
+        iv. [W3C](https://www.w3.org/) 
+        v. [WHATWG](https://whatwg.org/)
```
jensimmons, line 42: Shall we add the standardization groups that define image, video and audio formats? We get proposals every year for media formats. There may be good reasons to not include them (testability being one that's come up in the past), but it seems like "not the W3C, et al" is not a good one.

gsnedders: c.f. #589

foolip: I'd prefer to not do that now, let's update ahead of Interop 2025 instead, so that this matches what the proposal template said.

foolip: Oh, I see @jensimmons made a concrete proposal in https://github.com/web-platform-tests/interop/issues/589#issuecomment-1778208293. I'm OK with that since the work was already done. Jen, you said you created a suggested amendment, do you mean the comment itself, or do you have a suggested edit to this PR sitting as a draft somewhere?

2024/selection-process
stof, line 1: shouldn't the file have a `.md` extension so that github can render it instead of showing the raw markup ?

2024/selection-process.md
```diff
@@ -0,0 +1,139 @@
...
+4. Focus Area proposals must: 
+    A. Name a specific web technology. 
+        i. This can be a web technology that has not yet been implemented in any or all browser engines.
+        ii. Or this can be a web technology that has already been implemented in all browser engines, but those implementations have test failures that reveal a lack of compliance to the web standard, causing a lack of interoperability.
+    B. The feature must be defined in a sufficiently mature standards-track web specification from one the following organizations. Please provide a link to the web standard. 
+        i. [IETF](https://www.ietf.org/) (Standard Track)
+        ii. [Khronos Group](https://www.khronos.org/)
+        iii. [TC39](https://tc39.es/) (Stage 2+)
+        iv. [W3C](https://www.w3.org/) 
+        v. [WHATWG](https://whatwg.org/)
```
foolip, line 42: Suggested edit capturing the outcome of https://github.com/web-platform-tests/interop/issues/589:

```suggestion
        i. [Alliance for Open Media](https://aomedia.org/)
        ii. [IETF](https://www.ietf.org/) (Standard Track)
        iii. [ISO](https://www.iso.org/)
        iv. [Khronos Group](https://www.khronos.org/)
        v. [TC39](https://tc39.es/) (Stage 2+)
        vi. [W3C](https://www.w3.org/) 
        vii. [WebAssembly](https://webassembly.org/specs/) (Phase 3+)
        viii. [WHATWG](https://whatwg.org/)
```

@jensimmons if I got this right, can you apply it?

foolip: Also "or other similar industry web standards organizations"?

PR #631, opened 2024-02-04
Fix chapter sizes in interop 2024

atjn:
Some of the headers for chapters in interop 2024 are too small, making it look like the focus area is a subset of the previous focus area.


foolip (approved):
Thanks for spotting this!

PR #657, opened 2024-04-30
2025 proposal selection process

nairnandu:
Based on the retrospective and feedback from proposal authors, this is a draft of the 2025 proposal review process to be reviewed by the team

### DIFF
--- a/2025/selection-process.md
+++ b/2025/selection-process.md
@@ -1,4 +1,4 @@
-### Interop 2025 proposal selection process
+# Interop 2025 proposal selection process
 
 Interop 2025 is an effort to increase interoperability across browsers in key technical areas that are of high priority to web developers and end users. This effort is part of the [web-platform-tests](https://github.com/web-platform-tests/wpt) (WPT) project — an automated test suite for web standards — and run by a team of representatives from companies that make substantial contributions to browser rendering engines (including Apple, Bocoup, Google, Igalia, Microsoft, and Mozilla).
 
@@ -7,7 +7,7 @@ Please see the [Interop 2024 Dashboard](https://wpt.fyi/interop-2024) and [Inter
 Each iteration of the project starts with an open call for proposals, followed by a selection and prioritization process by each participating organization. Final decision on the inclusion of a proposal is made by [consensus](https://github.com/web-platform-tests/interop/blob/main/charter.md#:~:text=The%20team%20makes%20decisions%20based%20on%20consensus.%20A%20decision%20has%20consensus%20if%20it%20has%20support%20from%20at%20least%20two%20participating%20organizations%20and%20no%20opposition). For Interop 2025, the timeline and the selection process is as described below.
 
 
-### Timeline
+## Timeline
 
 This is the general timeline for the Interop 2025 proposal selection process. The dates are centered on the Interop team meeting (Thursdays). Please note that the timelines are subject to change.
 
@@ -18,10 +18,10 @@ This is the general timeline for the Interop 2025 proposal selection process. Th
 *   Scope of the Interop 2025 project will be published, the first half of February, 2025.
 
 
-### Proposal selection phases
+## Proposal selection phases
 
 
-#### Phase 1: Filtering (Oct 3rd to Oct 17th)
+### Phase 1: Filtering (Oct 3rd to Oct 17th)
 
 Interop meeting dates: Oct 3rd, Oct 10th, Oct 17th
 [Visibility] All decisions and results from this phase will be published on the Interop repo.
@@ -33,7 +33,7 @@ There are two steps to this phase:
 2. First round of feedback on proposals that do not meet the Interop project criteria would be provided to proposal authors on the Interop repo.
 
 
-#### Phase 2: Ranking (Oct 18th to Oct 31st)
+### Phase 2: Ranking (Oct 18th to Oct 31st)
 
 Interop meeting dates: Oct 24th, Oct 31st
 [Visibility] All decisions, made in Interop team meetings during this phase, will be confidential. However, organizations can choose to publish their own rankings, outside of the Interop program.
@@ -44,7 +44,7 @@ In this stage, the focus would be on organizations picking areas where there is 
 4. Any proposal that has at least 2 organizations expressing high priority and no objections, would be moved forward to the next phase of consensus building. Any organization that posts an objection would also outline the reasoning for it, to the Interop team. The ranking of proposals would be done asynchronously, ahead of the Interop meetings.
 
 
-#### Phase 3: Prioritization (Nov 1st to Dec 5th)
+### Phase 3: Prioritization (Nov 1st to Dec 5th)
 
 Interop meeting dates: Nov 7th, Nov 14th, Nov 21st, Dec 5th
 [Visibility] All decisions made in this phase will be confidential. 
@@ -56,7 +56,7 @@ In this stage, each organization will further prioritize proposals for a specifi
 7. Any proposal that has support from 2 organizations and does not have any objection, will be considered included in Interop 2025. At the end of this phase, the expectation is that the Interop team will have a first draft of proposals that are included in Interop 2025.
 
 
-#### Phase 4: Carryovers and Grouping (Dec 6th to Dec 19th)
+### Phase 4: Carryovers and Grouping (Dec 6th to Dec 19th)
 
 Interop meeting dates: Dec 12th and Dec 19th
 [Visibility] All decisions and results from this phase will be confidential.
@@ -68,7 +68,7 @@ In this stage, the Interop team will evaluate progress made in Interop 2024 focu
 10. This phase will also include decisions on inclusion of investigation effort proposals, by consensus.
 
 
-#### Phase 5: Final decisions (Jan 9th to Jan 30th)
+### Phase 5: Final decisions (Jan 9th to Jan 30th)
 
 Interop meeting dates: Jan 9th, Jan 16th, Jan 23rd, Jan 30th
 [Visibility] The final list of focus areas would be made public at launch
--- a/2025/selection-process.md
+++ b/2025/selection-process.md
@@ -24,6 +24,7 @@ This is the general timeline for the Interop 2025 proposal selection process. Th
 ### Phase 1: Filtering (Oct 3rd to Oct 17th)
 
 Interop meeting dates: Oct 3rd, Oct 10th, Oct 17th
+
 [Visibility] All decisions and results from this phase will be published on the Interop repo.
 
 In this stage, proposals that do not meet the criteria set by the Interop team will be filtered out. Feedback would be provided to the proposal authors to verify any additional data points that should be considered prior to exclusion. The Interop team will decide on the exclusion of proposals, based on consensus and proposal authors will be informed at the end of this phase. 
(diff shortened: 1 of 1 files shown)


2025/selection-process.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Phase 2: Ranking (Oct 18th to Oct 31st)
+
+Interop meeting dates: Oct 24th, Oct 31st
+
+[Visibility] All decisions, made in Interop team meetings during this phase, will be confidential. However, organizations can choose to publish their own priorities, for the Interop proposals, outside of the Interop program.
+
+In this stage, the focus would be on organizations picking areas where there is internal alignment on the specific feature or area being important enough to work on from a user/web developer perspective. Each organization will have different rubrics for prioritization. However, where applicable, organizations can choose to have any data used for prioritization be shared publicly.
+
+3. Each organization will select proposals for the prioritization discussion by putting a High/Low priority signal against each proposal where they have internal alignment to do the work. A low priority signal would indicate a willingness to support the proposal, if other organizations also deem it to be important. Organizations can also express any objections at this point.
```
dandclark, line 45: Is there a "neutral" vote option implied here as well, or is that effectively what the "Low priority" vote means?

For example, if an organization chooses not to place a vote on a given proposal, is that interpreted as if they'd voted "Low priority"? Or is there a difference in how these are recorded such that voting "Low priority" is a slightly more positive signal than "abstain"?

2025/selection-process.md
```diff
@@ -0,0 +1,81 @@
...
+
+Interop meeting dates: Nov 7th, Nov 14th, Nov 21st, Dec 5th
+
+[Visibility] All decisions made in this phase will be confidential. 
+
+In this stage, each organization will further prioritize proposals for a specific year (Interop 2025). This is more involved and a complex process where each organization weighs engineering priorities and available resources.
+
+5. From the list of proposals that get through to Phase 3, each organization will have an opportunity to further prioritize internally and build consensus within the Interop Team based on available signals.
+6. Organizations can also object to any proposal at this point.
+7. Any proposal that has support from 2 organizations and does not have an objection, will be considered included in Interop 2025. At the end of this phase, the expectation is that the Interop team will have a first draft of proposals that are included in Interop 2025.
```
dandclark, line 59: This voting process looks like it's effectively another iteration of the Phase 2 process, but without using the "High/Low priority" terminology.

Are there any actual procedural differences? If not perhaps the same terminology should be reused to prevent confusion.

2025/selection-process.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Phase 4: Carryovers and Grouping (Dec 6th to Dec 19th)
+
+Interop meeting dates: Dec 12th and Dec 19th
+
+[Visibility] All decisions and results from this phase will be confidential.
+
+In this stage, the Interop team will evaluate progress made in Interop 2024 focus areas and create a first draft of areas that will be carried over to Interop 2025. In addition, the proposals from phase 3 will be weighed against any carryover focus areas from Interop 2024. The outcome from this phase is a second draft of prioritized proposals for Interop 2025 (this will include new proposals and carryover focus areas) with a high-level grouping. This stage will include the steps below:
+
+8. Interop 2024 carryover evaluation - this will be done live in the Interop team meeting.
```
dandclark, line 70: This should be clearer about how carryover evaluation works. If there is no consensus on whether or not to carryover a given proposal, is it default-carryover or default-discard?

2025/selection-process.md
nairnandu: Makes sense to clarify. In a procedural sense, the High/Low signals from round 2 will carry forward into round 3. Organizations will have the option to change that signal in round 3.

jgraham:
Based on the feedback from last year, going forward I think we'll have more success if we're more successful at developing a shared understanding of value proposition of different proposals, and where we have room for discussion. Of course that won't always work, but I'd like the process to focus on finding those points of agreement and ensuring that as far as possible we end up with a shared consensus.

With that in mind, here's a different take on what the process could be:

1. After the proposal deadline each participating org shares a list of focus areas they'd like to champion. Those can correspond to one or more proposals. Where different orgs have similar focus area proposals they can work together out of band to develop one proposal, but in each case there should be one primary champion for later rounds. It's also OK to drop proposals at any stage if no one wants to champion it further.
2. Champions work on their focus areas, gathering evidence that the focus area represents an Interop priority (e.g. bug reports, developer feedback, positive platform impact on areas like a11y; we can discuss what people would like to see in advance). They are also responsible for ensuring the focus area has an identified set of tests. We do this in a shared space so that non-champions can provide feedback if desired (e.g. point to relevant standards-positions).
3. We have a fixed amount of f2f time (e.g. 15 minutes per participant over two meetings) to formally present the focus area proposals and explain why they're considered a priority for Interop.
4.  We have some time for further async adjustments e.g. if there's feedback like "we like X, but can't support it with subfeature Y, please remove that".
5. Each participant now buckets the final proposals as P1, P2, P3 or "veto". 
6. A successful outcome at this point looks like lots of proposals where orgs have given similar rankings, and few vetos. We sort by the buckets and pick focus area proposals with lots of P1/P2 rankings.

Some notes:
1. There is no formal elimination stage. It's up to participants to not champion proposals that don't meet the requirements to be accepted.
2. There's no limit on the number of proposals you can champion. It's up to individual orgs whether they'd prefer to put more time into fewer proposals or less time into more proposals. But at the presentation stage you get a fixed time to explain to others why you consider each proposal an Interop priority. This reflects the fact that time is the actual limited resource (both for decision making and implementation).

2025/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+Proposals with mixed rankings are ordered according to the number of
+P1 rankings, followed by the number of P2 rankings, followed by the
+number of P3 rankings. Starting from the top of the ordered list
+participants who ranked the proposal lower are given an opportunity to
+explain their concerns, and suggest any adjustments that would change
+their perspective. Proposal champions get final say on whether to
+adjust their proposals in response to this feedback.
+
+Following each meeting participants may adjust their rankings for the
+remaining proposals up or down in response to discussions, any agreed
```
foolip, line 160: As discussed in today's meeting, there's some ambiguity here about whether adjusting up and down is the _only_ way proposals end up being accepted or rejected, in particular as we approach the final decision on Dec 19th.

I think allowing for adjusted rankings in response to argument is good, and that will hopefully move a few things into the accepted state.

I would suggest that all remaining limbo proposals are individually decided by consensus, meaning two supporting and none opposed. This ensures that we have a decision on each proposal that reached this stage, and don't implicitly leave out the bottom N if we run out of time, for example.

2025/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+### Focus Area Ranking and Selection
+
+**Timeline**: December 5th, 12th, 19th
+
+Focus areas with any vetoes are eliminated.
+
+Proposals that have strong positive consensus (majority P1 rankings,
+no P3 rankings) are immediately adopted.  Proposals with a strong
+negative consensus (majority P3 rankings, no P1 ranking, or P1 only
+from the champion) are immediately dropped.
```
foolip, line 149: Does majority here mean 4 given that we have 6 participating organizations? The details here might change the number of proposals that are covered by this by a lot.

dandclark:
1741ff1b1c5c3c78b02a8a2da3284336b1e576aa removed all the language about which parts of the process are confidential and which are not. I think it's still important that this all be defined beforehand.

Microsoft's position is still that the Interop process should follow the model of web standards development by doing things in the open such that positions and discussions are public. But at the very least, organizations should be able to share which proposals they are championing and why, since revealing that information won't give full visibility into how the final votes and vetoes stack up.

foolip:
I agree it's important that any confidential parts are explicitly marked. Per our [charter](https://github.com/web-platform-tests/interop/blob/main/charter.md#governance) we operate in public except where a process document like this one says otherwise.

Google's preference is to operate in public, including our all or positions on specific proposals. If there is some confidentiality in the process, we think it's important to at the very least clarify that it only applies to sharing the positions of other organizations who want them to be confidential. This wording from the original revision of this PR would do the job:

> However, organizations can choose to publish their own priorities, for the Interop proposals, outside of the Interop program.

jgraham:
To be clear there's no intended change to the confidentiality compared to last year i.e. the details around proposal ranking and selection remain confidential to participants.

Of course there's not, and never has been, any restriction on commenting positively or negatively about specific web technologies in general.

I need to update the PR to make that clear.

However, the possibility of specifically allowing people to talk about the proposals they are championing is something I'd also considered, and would be worth discussing as a group.

2025/selection-process.md
jgraham: I've changed the wording to make it clear that it remains a judgement call, not an algorithm. Given that it's a new process it seems better to allow additional flexibility here.

2025/selection-process.md
jgraham: I've clarified at the end that we have to actually make a decision about any remaining proposals, that is they can't just be ignored. But I'm really hoping that we don't have to use that. If it's clear that everyone is "meh" about a proposal to the extent that we don't even discuss it that seems like a pretty clear sign that it's not important enough to be part of Interop.

2025/selection-process.md
```diff
@@ -0,0 +1,180 @@
...
+adjust their proposals in response to this feedback.
+
+Following each meeting participants may adjust their rankings for the
+remaining proposals up or down in response to discussions, any agreed
+changes to the proposals, or the overall composition of the project,
+and the process repeats in the following meeting.
+
+Before the end of the December 19th meeting there's a final decision
+on any proposals whithout a clear resolution formal call for
+consensus on the set of selected focus areas.
```
jgraham, line 165:
```suggestion
on any proposals whithout a clear resolution formal call for
consensus on the set of selected focus areas.
```
```suggestion
on any proposals without a clear resolution, and formal call for
consensus on the set of selected focus areas.
```

dandclark:
> Of course there's not, and never has been, any restriction on commenting positively or negatively about specific web technologies in general.
> 
> I need to update the PR to make that clear.
> 
> However, the possibility of specifically allowing people to talk about the proposals they are championing is something I'd also considered, and would be worth discussing as a group.

Last meeting I believe we were discussing these points but ran out of time. To help focus the discussion for next meeting I've pushed a suggestion for language that further clarifies the first point and opens up sharing of championed proposals. Let's dig into this more Thursday.

foolip:
Thank you @dandclark! I've included this in the agenda for tomorrow, https://github.com/web-platform-tests/interop/issues/674.

The wording you've suggested looks good to me, and is in line with Google's position/preference in https://github.com/web-platform-tests/interop/pull/657#issuecomment-2194230551.

2025/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+As in Interop 2023 and 2024, details of proposal selection and ranking
+remain participant confidential throughout the process. Final decisions
+will be published as the Interop team.
+
+Confidentiality does not prevent participating organizations from
+sharing their own priorities or roadmaps outside of the Interop program.
+
+### Prior to Proposal Submission
+
+**Timeline**: Before September 19th 2024
```
jgraham, line 51:
```suggestion
**Timeline**: Before September 12th 2024
```

2025/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+that are considered important when assessing proposals. These will be
+made available to proposal authors, and used later in the process when
+championing and assessing different proposals. However they are not
+binding on any specific participant i.e. participants are free to
+consider additional criteria (such as individual resource constraints)
+and weigh the shared criteria as they choose.
+
+### Proposal Submission
+
+**Timeline**: Before September 19th to October 3rd
```
jgraham, line 63:
```suggestion
**Timeline**: September 12th to October 3rd
```

Issue #674, opened 2024-07-10
Agenda for July 11th, 2024


foolip:
Here is the proposed agenda for the meeting on July 11th, 2024:

- Identify action items from signals discussion. Any updates to proposal templates?
- https://github.com/web-platform-tests/interop/pull/657

jensimmons:
Test change proposals? 

We'd like to get approval for this: https://github.com/web-platform-tests/interop/issues/672

foolip:
Brief notes from yesterday's meeting:

Attendees: @bkardell, @jgraham, @dandclark, @foolip, @stubbornella, @nt1m, @meyerweb 

- Consensus to exclude XR tests in https://github.com/web-platform-tests/interop/issues/672
- Follow up from signals discussion. Any changes needed on proposal templates?
  - @stubbornella: Main thing for us was a shared understanding. Maybe giving some examples would be helpful.
  - @bkardell: There would be value in writing something (anywhere) that we can point to.
  - @jgraham: The current process proposal says that we’ll write something to inform proposal authors. The templates didn’t work very well last year, so should we cut back a bit on those? Instead, we have “please read these guidelines” which mentions signals that we’ll consider.
  - @stubbornella: That makes sense to me. We’re going to validate signals anyway, so having proposers fill in the template doesn’t reduce our work. The champion for each proposal can do it, and would anyway.
  - @dandclark: I worry that people won’t follow links and read the docs. But I don’t object to trying this.
  - @jgraham: You’re probably right, most people didn’t read it when it was in the template even. It’s a bit about having something to point to to explain what we’re taking into consideration. Not just for proposers but anyone who’s hoping for a certain outcome.
  - @foolip: Agree this all makes sense. Too much template in past years.
  - @jgraham: I think we need (1) in public, things that we’re looking for in a proposal and (2) internally, a shared spreadsheet for the objective signals.
  - @stubbornella: Can we volunteer you for the first bit?
  - @jgraham: I can start a PR which we can get feedback on.
  - @stubbornella: I’m happy to look at the shared spreadsheet. There’s a question of whether we make it public. It’s hard to judge and I don’t want web developers to be discouraged.
  - @jgraham: I think it should be confidential.
  - @dandclark: Sharing it would be useful to learn if we’ve made a mistake, and it wouldn’t be attributed to any of us.
  - @stubbornella: The part that wouldn’t work in public is if I survey our engineers. I’ll need to think that over.
[...]

PR #675, opened 2024-07-11
Dandclark/2025 process share championed proposals

dandclark:
Allow organizations to share publicly the list of proposals that they are championing. This was discussed in the [July 11 meeting](https://github.com/web-platform-tests/interop/issues/674) but did not reach consensus.

Change the wording to clarify that what is shared is the actual list of proposals the org is selected as champion for, not the list of proposals they were initially interested in championing.


foolip (approved):
Google would like to see this as a very small step to more transparency in the process, and preferably a return to complete transparency as in the Interop 2022 planning process.

PR #680, opened 2024-07-25
Add Interop 2025 proposal templates

foolip:


foolip:
This shouldn't be merged until we open up for proposals, since it will show up as an option in https://github.com/web-platform-tests/interop/issues/new/choose.

PR #694, opened 2024-09-06
Update selection-process.md

nairnandu:
Proposed changes to the dates for Interop 2025 process

### DIFF
--- a/2025/selection-process.md
+++ b/2025/selection-process.md
@@ -28,7 +28,7 @@ This is the general timeline for the Interop 2025 proposal selection
 process. The dates are centered on the Interop team meeting (Thursdays).
 Please note that the timelines are subject to change.
 
-*   Proposal submission window (3 weeks): Sept 17th, 2024 to Oct 7th, 2024
+*   Proposal submission window (~3 weeks): Sept 17th, 2024 to Oct 9th, 2024
 *   Proposal selection: Before December 19th 2024
 *   Scope of the Interop 2025 project published: the first half of February, 2025.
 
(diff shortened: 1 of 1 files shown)


2025/selection-process.md
```diff
@@ -60,7 +60,7 @@ and weigh the shared criteria as they choose.
 
 ### Proposal Submission
 
-**Timeline**: September 12th to October 3rd
+**Timeline**: September 17th to October 7th
```
jgraham, line 63: Why don't we make this October 10th (or 9th if it should unambiguously be before the meeting)?

2025/selection-process.md
```diff
@@ -82,12 +82,12 @@ investigation efforts.
 
 ### Champion Selection
 
-**Timeline**: October 3rd, 10th, 17th
+**Timeline**: October 17th and 24th
```
jgraham, line 85: Why not 10th, 17th and 24th?

jgraham (commented):
So the overall effect of this is to push everything back a week until initial (internal) ranking? That seems OK to me, but we should make sure everyone is on board.

tantek (approved):
LGTM with changes suggested by @jgraham

2025/selection-process.md
nairnandu: Oct 9th makes sense to me

2025/selection-process.md
nairnandu: 10th was excluded to give some buffer, in case there is a late flurry of proposals and teams needed time to review. That said, since there is no expectation defined for the 10th, I don't see any harm in adding it back in.

PR #862, opened 2024-10-09
Close Interop 2025's call for proposals

nairnandu:


dandclark (approved):

jgraham (approved):

sandstrom:
@dandclark Is it too late to propose more items?

My suggestion is better cross-browser support for `<datalist>`. The MDN docs has some good examples:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

In short, some things that doesn't work well across browser are:

- No date and time support in Firefox and Safari.
- Broken range support in Safari (or doesn't work like Chrome at least, no snapping).
- Mobile support is spotty (broken in iOS currently, suggestions when typing has started is so-so)
- Inconsistency across browsers in whether the options are shown immediately, or after typing has started.
- Not easy to style the UI of the suggestions (one could argue that's a feature too though). Firefox doesn't 

### Argument

This is a very powerful primitive which has the potential for autocomplete experience for users with zero Javascript. However, at the current level of polish it's hard to justify using it.

### Resources

- https://caniuse.com/datalist
- https://www.peterbe.com/plog/datalist-looks-great-on-mobile-devices
- https://chriswarrick.com/blog/2020/02/09/when-html-is-not-enough-a-tale-of-the-datalist-element/

PR #919, opened 2025-02-06
Focus area descriptions

captainbrosset:


DanielRyanSmith:
This will need to be merged if we want it available on the dashboard for the release date. Has there been any movement on this?

captainbrosset:
The plan is, I think, to merge it on launch day, moments before the dashboard goes out.
I've not received any feedback so far on the PR. Note that there had been some healthy amount of feedback on the document which this PR comes from. So this has already gone through one round of review.

jgraham:
> WebAssembly -> WebAssembly Additions

Without wanting to bikeshed, we have fairly often chosen more generic names over less generic ones (c.f. "WebRTC" which isn't the entire WebRTC API, "Layout" which isn't all of CSS, etc.). I think "additions" in particular doesn't add much value in the name since many focus areas are about adding new functionality to an existing broad area of the platform and ends up sounding a bit like the feature is to do with arithmetic.

2025/README.md
```diff
@@ -1,18 +1,396 @@
...
+  * [Navigation API](#navigation-api)
+  * [Pointer and Mouse events](#pointer-and-mouse-events)
+  * [Remove mutation events](#remove-mutation-events)
+  * [`@scope`](#scope)
+  * [`scrollend` event](#scrollend)
+  * [Storage Access API](#storage-access-api)
+  * [`text-decoration`](#text-decoration)
+  * [URLPattern](#urlpattern)
+  * [View Transitions](#view-transitions)
+  * [WebAssembly](#webassembly)
```
nt1m, line 29:
```suggestion
  * [WebAssembly Extensions](#webassembly)
```

It's only about two specific features AFAIK, not the whole of WASM:

* JS string builtins: to make the WebAssembly builtin string functions mirror a subset of the JavaScript String API so it can be callable without JavaScript glue code.
* Resizable buffers integration: to integrate WebAssembly into JS code that uses resizable buffers.

WabAssembly JS integration might work too? I'm open to different wording.

nt1m:
> > WebAssembly -> WebAssembly Additions
> 
> Without wanting to bikeshed, we have fairly often chosen more generic names over less generic ones (c.f. "WebRTC" which isn't the entire WebRTC API, "Layout" which isn't all of CSS, etc.). I think "additions" in particular doesn't add much value in the name since many focus areas are about adding new functionality to an existing broad area of the platform and ends up sounding a bit like the feature is to do with arithmetic.

I think there's a balance to reach, we don't use "Color" for "Relative Color Syntax" for instance. Perhaps something like WebAssembly JS integration or whatever name is appropriate.

I'd probably suggest a longer name for WebRTC too, but I don't think there's a more detailed name that makes sense that isn't too long for the dashboard. Ditto for Layout.

Issue #755, opened 2024-09-26
Dividing Scalars in calc()


nmn:
### Description

Calc currently doesn’t allow dividing one length by another.

Change this to allow dividing any pair of values than can be added and return a unit-less number. This number can then be used in various different scenarios.

A specification for this exists for many years. It’s possible to do this already using `atan2()` function with slightly buggy results in some browsers.

There are some very useful use cases where this would dramatically simplify layout code which would otherwise need a large number of container queries.

### Specification

W3C: https://www.w3.org/TR/css-values-4/#calc-type-checking

### Tests

[css/css-values/getComputedStyle-calc-mixed-units-003.html](https://wpt.fyi/results/css/css-values/getComputedStyle-calc-mixed-units-003.html)
[css/css-values/calc-in-media-queries-with-mixed-units.html](https://wpt.fyi/results/css/css-values/calc-in-media-queries-with-mixed-units.html)

### Browser bug reports

WebKit: https://bugs.webkit.org/show_bug.cgi?id=255280
Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1827404
Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=1432187 and https://bugs.chromium.org/p/chromium/issues/detail?id=1213338

### Additional Signals

This is a purely additive feature that should not cause any regressions.

* https://twitter.com/anatudor/status/1537881090223853568

nt1m:
See last year's proposal for more context: https://github.com/web-platform-tests/interop/issues/513

nmn:
Let me quickly provide my own use-case for this:

I want a horizontally scrollable list of cards where the cards fit the available space perfectly regardless of the available space. This a fluid grid, but with all the rows combined on a single row.

This is how this ability in `calc` would make that possible:

```css
width: calc(100% / round(down, 100% / var(--min-card-width)))

/* or with small parts */
--num-cards-that-fit: round(down, 100% / var(--min-card-width));
width: calc(100% / var(--num-cards-that-fit));
```

Without this `calc` function, we can use a long list of `@container` queries instead.
In production, I use a huge amount of nested `clamp()` function calls with the "fab-four technique". It works today, but it's ugly, unmaintainable code. Dividing lengths with `calc` would make it clean and obvious.

scottkellum:
Great to see this here thanks for the submission @nmn 👍🏻 I submitted this last year and there has been progress, with Webkit supporting this in Safari TP 203.

[...]

PR #960, opened 2025-05-19
First draft of selection-process.md

nairnandu:
First draft for review and comments. This is a copy of last year's process, with date changes.

### DIFF
--- a/2026/selection-process.md
+++ b/2026/selection-process.md
@@ -44,7 +44,8 @@ remain participant confidential throughout the process. Final decisions
 will be published as the Interop team.
 
 Confidentiality does not prevent participating organizations from
-sharing their own priorities or roadmaps outside of the Interop program.
+sharing their own priorities or roadmaps, as distinct from positions in the
+proposal selection process.
 
 ### Prior to Proposal Submission
 
--- a/2026/selection-process.md
+++ b/2026/selection-process.md
@@ -1,0 +1,183 @@
+# Interop 2026 proposal selection process
+
+Interop 2026 is an effort to increase interoperability across browsers
+in key technical areas that are of high priority to web developers and
+end users. This effort is part of the
+[web-platform-tests](https://github.com/web-platform-tests/wpt) (WPT)
+project — an automated test suite for web standards — and run by a
+team of representatives from companies that make substantial
+contributions to browser rendering engines (including Apple, Bocoup,
+Google, Igalia, Microsoft, and Mozilla).
+
+Please see the [Interop 2025 Dashboard](https://wpt.fyi/interop-2025)
+and [Interop 2025
+README](https://github.com/web-platform-tests/interop/blob/main/2025/README.md)
+for the current iteration of the project.
+
+Each iteration of the project starts with an open call for proposals,
+followed by a selection and prioritization process by each
+participating organization. Final decision on the inclusion of a
+proposal is made by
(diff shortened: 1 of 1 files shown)


2026/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+Interop participants propose focus areas, consisting of one or more
+proposals, that they would like to champion. Proposals will initially
+be shared in time for the meeting on October 16th, and champions
+finalised on October 23rd.
+
+Where more than one participant wants to champion the same focus area
+proposal a single champion must be nominated for the remainder of the
+process.
+
+Champions and seconds (i.e. organizations that would champion a focus
```
dandclark, line 96: Did we actually use the concept of "seconds" last year? I don't remember those getting assigned.
Unless we expect that concept to be useful this year, we should consider dropping it.

2026/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+Each participant gets 15 minutes of meeting time to present the
+proposals they are championing, and make the case that they should be
+considered a priority for inclusion in Interop 2026.
+
+### Focus Area Proposal Ranking
+
+**Timeline**: Before December 4th
+
+Participants rank each proposal as P1 (highest priority) to P3 (lowest
+priority), or veto.
```
dandclark, line 135:
```suggestion
priority), or veto. Ranks are assigned to the individual proposals, not the overall focus areas.
```
This was a point of confusion for Interop2025 so we should try to clarify it here.

I could also live with assigning ranks to the overall focus areas, but I thought there was a lot of useful signal from folks having ranked individual proposals. It's easier to spot opportunities to move things around to make overall focus areas more viable.

2026/selection-process.md
```diff
@@ -0,0 +1,183 @@
...
+understanding of the merits of each proposal, and converge on a
+decision about the most impactful areas in which we can improve
+interop on the web over 2026.
+
+As in Interop 2024 and 2025, details of proposal selection and ranking
+remain participant confidential throughout the process. Final decisions
+will be published as the Interop team.
+
+Confidentiality does not prevent participating organizations from
+sharing their own priorities or roadmaps outside of the Interop program.
```
foolip, line 47: After we launched Interop 2025 I commented on some proposals to give additional feedback, for example in https://github.com/web-platform-tests/interop/issues/755#issuecomment-2674354422.

I wasn't sure if commenting about our plans for https://crbug.com/40768696 on the proposal itself would be OK, so I'd like to clarify that here. The simplest edit I'd like to suggest is to remove the "outside the Interop program" part:

```suggestion
sharing their own priorities or roadmaps.
```

In other words, it's OK to share such information both during the process and to follow up later to say that a bug has been fixed or a feature shipped, for example. This information is often discoverable from a bug or something linked from the proposal anyway.

foolip: I understand the concern, and think it comes down to the expectation that positions in the proposal selection process aren't revealed with a wink and a nudge. In past years we've proposed that confidentiality shouldn't apply to our own Interop positions, and that's still a change I'd welcome, but that's not the intention here.

Wording is hard, but the best I've been able to come up with is something like this:

> Confidentiality does not prevent participating organizations from sharing their own priorities or roadmaps, as distinct from positions in the proposal selection process.

I'd be OK with or without the extra clause, to me it's clear in either case that trying to find loopholes for confidentiality will not be well received and would be a breach of trust.

foolip: This was discussed in the meeting today. @jgraham preferred the additional clause, so as an edit it would be this:

```suggestion
sharing their own priorities or roadmaps, as distinct from positions in the
proposal selection process.
```

@stubbornella can you take a look?

2026/selection-process.md
nairnandu: Updated.

2026/selection-process.md
nairnandu: updated

PR #978, opened 2025-07-17
GitHub action to update interop proposals with web-features-explorer data

captainbrosset:
This is a PR for a proposal I made on a recent interop team meeting, also based on a discussion with @jgraham, and thanks to feedback I got from multiple people at that meeting.

The PR consists of a GitHub action which runs when a new issue is created, and when an existing issue is edited.
The action runs a NodeJS script (in the new `/scripts/` folder), which reads the issue body and attempts to find web-features that correspond to various URLs found in the body.

This can obviously be improved, but I wanted to open this PR early for a first round of feedback/review.

I don't know of a way to test GitHub actions without pushing changes to the main branch of a repo though, so I tested this on my own repo here: https://github.com/captainbrosset/interop-proposals-bot, and you can see a few issues which I experimented with: https://github.com/captainbrosset/interop-proposals-bot/issues

To test this further, feel free to create more issues on my repo.

So far, I mainly tested this in "happy path" scenarios, for example with a spec URL that actually matched a web-features entry. However, there are many focus area proposals from last year which don't return any web-features because they either don't have data we can use, or because the spec URL differs slightly.
So, one thing we should do here is test this with a bigger variety of existing issues and improve the web-features detection code accordingly.

### DIFF
--- a/scripts/identify-web-features.js
+++ b/scripts/identify-web-features.js
@@ -32,96 +32,107 @@ async function getReferencedIssue() {
   return response.data;
 }
 
-// Identify web-features based on spec URLs in the issue body.
-function gatherFeaturesFromSpecUrls(issueBody, gatheredFeatures) {
-  // Find URLs in the issue body.
+function gatherUrlsFromIssue(issueBody) {
   const urls = issueBody.match(/https?:\/\/[^)\s]+/g) || [];
+  return urls.map(url => new URL(url));
+}
 
-  // Filter out known non-spec URLs.
-  const specUrls = urls
-    .filter(url => {
-      // Filter out known non-spec URLs.
-      return !url.includes("bugzilla") &&
-        !url.includes("github.com") &&
-        !url.includes("webkit.org") &&
-        !url.includes("developer.mozilla.org") &&
-        !url.includes("developer.chrome.com") &&
-        !url.includes("wpt.fyi") &&
-        !url.includes("css-tricks") &&
-        !url.includes("webstatus.dev") &&
-        !url.includes("learn.microsoft.com") &&
-        !url.includes("chromium.org");
-    }).map(url => {
-      // Separate the # from the URL if it exists.
-      return [url, url.split("#")[0]];
-    });
+// Identify web-features based on spec URLs in the issue body.
+function gatherFeaturesFromSpecUrls(urls) {
+  const gatheredFeatures = new Set();
 
-  // Iterate over web features to find spec URL matches.
-  for (const [candidateUrl, candidateUrlNoAnchor] of specUrls) {
+  for (const url of urls) {
     for (const id in features) {
       const feature = features[id];
-      const featureSpecs = Array.isArray(feature.spec) ? feature.spec : [feature.spec];
-      if (featureSpecs.some(spec => spec === candidateUrl || spec === candidateUrlNoAnchor)) {
+      const specUrls = (Array.isArray(feature.spec) ? feature.spec : [feature.spec]).map(url => new URL(url));
+
+      if (specUrls.some(specUrl => {
+        return specUrl.hostname === url.hostname &&
+          specUrl.pathname === url.pathname &&
+          (specUrl.hash ? specUrl.hash === url.hash : true);
+      })) {
         gatheredFeatures.add(id)
       }
     }
   }
+
+  return gatheredFeatures;
 }
 
 // Identify web-features based on explorer URLs in the issue body.
-function gatherFeaturesFromExplorerUrls(issueBody, gatheredFeatures) {
-  // https://web-platform-dx.github.io/web-features-explorer/features/<feature-id>
-  let urls = issueBody.match(/https?:\/\/web-platform-dx.github.io\/web-features-explorer\/features\/[a-z0-9-]+/g) || [];
-  urls = urls.map(url => url.split("#")[0]);
+function gatherFeaturesFromExplorerUrls(urls) {
+  const gatheredFeatures = new Set();
 
   for (const url of urls) {
-    // Extract the feature ID from the URL.
-    const match = url.match(/features\/([a-z0-9-]+)/);
-    if (match && match[1]) {
-      const id = match[1];
-      const feature = features[id];
-      if (feature) {
-        gatheredFeatures.add(id);
-      }
+    if (url.hostname !== "web-platform-dx.github.io" || !url.pathname.startsWith("/web-features-explorer/features/")) {
+      continue;
+    }
+
+    const candidateId = url.pathname.substring(url.pathname.indexOf("features/") + 9).replace("/", "").replace(".json", "");
+    if (features[candidateId]) {
+      gatheredFeatures.add(candidateId);
     }
   }
+
+  return gatheredFeatures;
 }
 
 // Identify web-features based on WPT URLs in the issue body.
-function gatherFeaturesFromWPTUrls(issueBody, gatheredFeatures) {
-  // https://wpt.fyi/results/?q=feature:<feature-id>
-  let urls = issueBody.match(/https?:\/\/wpt\.fyi\/results\/\?q=feature:[a-z0-9-]+/g) || [];
-  urls = urls.map(url => url.split("#")[0]);
+function gatherFeaturesFromWPTUrls(urls) {
+  const gatheredFeatures = new Set();
 
   for (const url of urls) {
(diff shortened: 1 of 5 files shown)


scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+  return response.data;
+}
+
+// Identify web-features based on spec URLs in the issue body.
+function gatherFeaturesFromSpecUrls(issueBody, gatheredFeatures) {
+  // Find URLs in the issue body.
+  const urls = issueBody.match(/https?:\/\/[^)\s]+/g) || [];
+
+  // Filter out known non-spec URLs.
+  const specUrls = urls
```
jgraham, line 41: I'd put the URLs through the URL parser at this point rather than working with the string representation.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+
+async function getReferencedIssue() {
+  const response = await octokit.request(`GET /repos/${argv.repo}/issues/${argv.number}`,);
+  return response.data;
+}
+
+// Identify web-features based on spec URLs in the issue body.
+function gatherFeaturesFromSpecUrls(issueBody, gatheredFeatures) {
+  // Find URLs in the issue body.
+  const urls = issueBody.match(/https?:\/\/[^)\s]+/g) || [];
```
jgraham, line 38: It feels like a better factoring here would be to have a function that returns URL objects for all the URLs in the issue and then have some functions that try to extract features from the list of URLs, rather than having each function do its own parsing.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+      return [url, url.split("#")[0]];
+    });
+
+  // Iterate over web features to find spec URL matches.
+  for (const [candidateUrl, candidateUrlNoAnchor] of specUrls) {
+    for (const id in features) {
+      const feature = features[id];
+      const featureSpecs = Array.isArray(feature.spec) ? feature.spec : [feature.spec];
+      if (featureSpecs.some(spec => spec === candidateUrl || spec === candidateUrlNoAnchor)) {
+        gatheredFeatures.add(id)
```
jgraham, line 65: In general I'd prefer that these functions return a set of possible features, rather than editing a set in-place.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+    }).map(url => {
+      // Separate the # from the URL if it exists.
+      return [url, url.split("#")[0]];
+    });
+
+  // Iterate over web features to find spec URL matches.
+  for (const [candidateUrl, candidateUrlNoAnchor] of specUrls) {
+    for (const id in features) {
+      const feature = features[id];
+      const featureSpecs = Array.isArray(feature.spec) ? feature.spec : [feature.spec];
```
jgraham, line 63: Again parsing the URL and comparing host and path and fragment if present is probably better than string comparisons.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+  }
+}
+
+function printBaselineStatus(feature) {
+  if (feature.status && feature.status.baseline === "high") {
+    return "Widely Available";
+  } else if (feature.status && feature.status.baseline === "low") {
+    return "Newly Available";
+  }
+  return "Limited Availability";
```
jgraham, line 148: There might also be totally unimplemented features, right?

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+  try {
+    const response = await fetch(`https://web-platform-dx.github.io/web-features-explorer/features/${id}.json`);
+    return await response.json();
+  } catch (error) {
+    console.error(`Error fetching the feature data for ${id}:`, error);
+    return null;
+  }
+}
+
+function printBaselineStatus(feature) {
```
jgraham, line 142: Extremely nitpicky, but I don't love "print" in the names of functions that don't actually do IO.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+    console.log("Could not find any matching features the issue body.");
+    return;
+  }
+
+  console.log(`Found ${features.length} matching feature(s):`);
+  console.log(features.map(f => `- ${f.id}`).join("\n"));
+
+  let content = "_This comment was automatically generated based on the information you provided. Please don't edit it._\n\n";
+  content += `Below is additional information about the web feature${features.length > 1 ? "s" : ""} (from the [web-features project](https://github.com/web-platform-dx/web-features/)) which ${features.length > 1 ? "are" : "is"} referenced in your proposal.`;
+  content += `This information might help motivate your focus area proposal.\n\n`;
```
jgraham, line 282: I don't think we need this line; I think the point of this is to tell _us_ which features are being proposed. I think we need something informing the user what to do if the output is _wrong_, especially if no features are found (which we should also handle below).

Something like:
"If these web features do not accurately correspond to your proposal, please add a comment below, or update your initial comment to include the string `web-features: <feature name>`"

(that assumes that we'll have some explicit web-feature id that overrides any other fields).

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+    if (features.length > 1) {
+      content += `<details>\n`;
+      content += `<summary>${escapeFeatureName(feature)}</summary>\n\n`;
+      content += featureContent;
+      content += `</details>\n\n`;
+    } else {
+      content += featureContent;
+    }
+  }
+
```
jgraham, line 296: We need to handle the case where no features were found, with a helpful message like
"No web-features found in the proposal. If your proposal doesn't correspond to a web feature, that is fine. Otherwise please add the string `web-feature: <feature name>` to the initial comment to link your proposal to the corresponding feature".

scripts/identify-web-features.js
```diff
@@ -0,0 +1,303 @@
...
+// Given a GitHub issue, find the web-features that are referenced in the issue body.
+// For now, we only look for URLs that match the web-features specifications.
+function findFeaturesInIssue(issue) {
+  const features = new Set();
+  
+  gatherFeaturesFromSpecUrls(issue.body, features);
+  gatherFeaturesFromExplorerUrls(issue.body, features);
+  gatherFeaturesFromWPTUrls(issue.body, features);
+  // TODO: Match on other things too:
+  // - web-features ID in the body of the issue, which might be there since the issue template asks for it.
```
jgraham, line 118: I think this one is important to support and should _override_ features we get from other sources. That means that the author (or whoever can edit the topmost comment) will be in full control.

jgraham (commented):
Thanks for doing this! It generally looks good, but a few code-level comments and a few comments about more important functional issues that I think we should address.

scripts/identify-web-features.js
captainbrosset: Yes, but Limited Availability currently encompasses them too.

scripts/identify-web-features.js
```diff
@@ -0,0 +1,320 @@
...
+  str += `* **ID:** ${feature.id}\n`;
+  str += `* **Name:** ${escapeFeatureName(feature)}\n`;
+  str += `* **Description:** ${feature.description_html}\n`;
+  str += `* **Baseline status:** ${getBaselineStatusAsMarkdown(feature)}\n`;
+  str += getDocsAsMarkdown(feature);
+  str += getStandardPositionsAsMarkdown(feature);
+  str += getUseCounterAsMarkdown(feature);
+  str += getSurveysAsMarkdown(feature);
+  str += getPreviousInteropsAsMarkdown(feature);
+  str += getWPTLinkAsMarkdown(feature);
```
foolip, line 238: Do you have a issue where the WPT link is added and working? It seems like everything is there in https://github.com/captainbrosset/interop-proposals-bot/issues/1 for it to be added, but it's not. Maybe there's a bug here?

captainbrosset: Good catch, fixed it now.

captainbrosset:
@jgraham @foolip I've just reviewed this again, fixed a couple minor problems, and tested it on my repo. I think this is about ready to merge if you agree.
As discussed at yesterday's Interop team meeting, we can always disable the workflow if we find issues later.

jgraham (approved):

captainbrosset:
We decided to merge this, to try it out, and if it causes problems, we can turn it off again.

Issue #1012, opened 2025-09-05
XSL Transformations (XSLT) Version 3.0


dbushell:
### Description

Web browsers support **XSLT 1.0** but there are issues (discussed in the _"Additional Signals"_ links below). This proposal is to suggest fixing security issues and implementing the **XSLT 3.0** spec allowing new features for web developers.

### Specification

https://www.w3.org/TR/xslt-30/

### web-feature

xslt

### Test Links

* XSLT 3.0 Test Suite https://github.com/w3c/xslt30-test/

### Additional Signals

* Should the web platform adopt XSLT 3.0? https://github.com/whatwg/html/issues/11578
* Should we remove XSLT from the web platform? https://github.com/whatwg/html/issues/11523

svgeesus:
The automatically discovered web feature `xslt` is for XSLT 1.0. So not the same as this proposal for XSLT 3.0.

wpt-interop:
Thank you for proposing XSL Transformations (XSLT) Version 3.0 for Interop 2026.

We want to let you know it was not selected to be part of Interop this year.

On behalf of the entire Interop team, thank you for submitting it for consideration. We got many more proposals than we could include in this year's project, necessitating some difficult choices.  Please note this should not be taken as a comment on the technology as a whole, or our willingness to consider it for Interop in the future.  We appreciate the work you put into your proposal, and would welcome your participation in future rounds of Interop.

_Posted on behalf of the Interop team._

PR #1015, opened 2025-09-05
Lower case web-features IDs

captainbrosset:
https://github.com/web-platform-tests/interop/issues/1012 is not getting recognized by the web-features bot because the id XSLT is upper case. All web-features IDs are lowercase.

### DIFF
--- a/scripts/identify-web-features.js
+++ b/scripts/identify-web-features.js
@@ -70,7 +70,7 @@ function gatherFeaturesFromExplorerUrls(urls) {
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


foolip:
I edited https://github.com/web-platform-tests/interop/issues/1012 as that was the only way to re-run the workflow. So the immediate problem is fixed.

foolip:
Oops, we never merged this. @captainbrosset would bad things happen if we merge this now?

captainbrosset:
I don't think anything bad would happen _now_, because the bot won't be running now that the proposal window has closed. But I'd prefer it if we didn't merge and re-visited this next year instead.

PR #1216, opened 2025-11-03
Upgrade to web-features 3.6.0

captainbrosset:
The web-features bot which we used during the proposal window uses an outdated version of the web-features data library.
This PR upgrades to 3.6.0, which is the latest version of the library, and fixes the code to account for the breaking changes that came with version 3.

### DIFF
--- a/scripts/identify-web-features.js
+++ b/scripts/identify-web-features.js
@@ -24,6 +24,15 @@ const argv = yargs(process.argv)
 
 const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
 
+function getFeature(id) {
+  const feature = features[id];
+  if (!feature || feature.kind !== "feature") {
+    return null;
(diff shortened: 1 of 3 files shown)


scripts/identify-web-features.js
```diff
@@ -24,6 +24,15 @@ const argv = yargs(process.argv)
 
 const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
 
+function getFeature(id) {
+  const feature = features[id];
+  if (!feature || feature.kind !== "feature") {
```
foolip: Do you see any benefit from handling moved features? It could be done by returning `[id, data]` perhaps, so that callers that use the ID could use the new one.

foolip (approved):
This looks like it'll work, handling renames would be nice, but up to you if you feel like it :)

PR #1263, opened 2026-02-20
Mappings for Interop 2026 focus areas to web-features

captainbrosset:
This adds the mappings between the Interop 20206 focus areas and web-features.

New web-features that were introduced to better match Interop focus areas:

* `wasm-jspi`, via https://github.com/web-platform-dx/web-features/pull/3791
* `navigation-precommit-handlers`, via https://github.com/web-platform-dx/web-features/pull/3792
* `fetch-formdata`, via https://github.com/web-platform-dx/web-features/pull/3813

Decisions for what _not_ to map:

* The remaining parts of the Fetch focus area are mime-type support and Range header support. Both of these are tracked in BCD in parts that aren't necessarily related to the Fetch API. So we can't really establish a useful mapping here.
* Timing of scroll events relative to animation events is not represented as a web-feature, and is more of a bug anyway.

### DIFF
--- a/web-features.json
+++ b/web-features.json
@@ -344,7 +344,10 @@
       "scroll-snap"
    ],
     "interop-2026-shape": [
-      "shape-function"
+      "shape-function",
+      "shapes",
+      "path-shape",
+      "rect-xywx"
     ],
     "interop-2026-view-transitions": [
       "view-transitions",
--- a/web-features.json
+++ b/web-features.json
@@ -318,21 +318,21 @@
       "popover-hint"
     ],
     "interop-2026-fetch": [
-      "fetch-request-streams"
-      // Missing other parts
+      "fetch-request-streams",
+      "fetch-formdata"
     ],
     "interop-2026-indexeddb": [
       "getallrecords"
     ],
     "interop-2026-jspi-for-wasm": [
-      "wasm-jspi" // Wait for this PR: https://github.com/web-platform-dx/web-features/pull/3791
+      "wasm-jspi"
     ],
     "interop-2026-media-pseudo-classes": [
       "media-pseudos"
     ],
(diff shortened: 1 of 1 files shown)


captainbrosset:
@foolip and/or @jgraham please take a look and let me know of any changes you'd like to make.

web-features.json
```diff
@@ -295,5 +295,77 @@
     "interop-2025-writingmodes": [
       "writing-mode"
     ]
+  },
+  "2026": {
+    "interop-2025-anchor-positioning": [
+      "anchor-positioning"
+    ],
+    "interop-2026-attr": [
+      "attr"
```
foolip, line 304: Yep, as opposed to attr-contents.

web-features.json
```diff
@@ -295,5 +295,77 @@
...
+      "scoped-custom-element-registries"
+    ],
+    "interop-2026-scroll-driven-animations": [
+      "scroll-driven-animations"
+    ],
+    "interop-2026-scroll-snap": [
+      "scroll-snap"
+   ],
+    "interop-2026-shape": [
+      "shape-function"
```
foolip: Aren't there more things to include per your 'Name and description of the "CSS shape()" focus area' email?

foolip (approved):
Approving so you can land without waiting for another round of review.

web-features.json
captainbrosset: Ah, yes, you're right. I think I wrote this PR before that email, and then forgot to update the PR. Done now. Thank you.

captainbrosset:
Thank you @foolip. I've addressed your comment about shapes. I can't merge on this repo, so will need you or somebody else to do it please.

foolip (approved):

PR #1318, opened 2026-08-11
web-features bot improvements for interop 2027

### LINKED ISSUE #1317: Improve the web-features finder bot for new proposals
Last year, we used the [Identify Web Features in Issues](https://github.com/web-platform-tests/interop/actions/workflows/identify-web-features.yml) action ([workflow source](https://github.com/web-platform-tests/interop/blob/main/.github/workflows/identify-web-features.yml), [script source](https://github.com/web-platform-tests/interop/blob/main/scripts/identify-web-features.js)) to relate focus area proposals to one or more web-features.

This was useful, and we want to do it again for Interop 2027.

This issue is here to discuss about the things that are worth improving in the action before we run it again this year.

I reviewed most of the focus area proposals from last year, searching for cases where the action didn't work quite right. Here are my findings of the common issues:

- Missing web-features ID (or other piece of information we use to identify features) in the proposal:

  - Sometimes, the author didn't know or bother.
  
  - Sometimes, The proposal was not a web-feature (e.g. a list of WPT tests from a particular area, or an investigation area).
  
  - Often, the web-features entry didn't exist yet, so the author couldn't add it.
  
  - Sometimes, the author, or somebody else, added the web-features ID later, after the proposal was created.

- Action found an incorrect feature:

  - Sometimes the proposal was about a very specific addition to an existing feature, but the action found the more general, parent feature instead.

  - Sometimes the author entered the wrong feature ID, because we didn't have a feature for their proposal yet:
    - In https://github.com/web-platform-tests/interop/issues/1012, `xslt` was found for a proposal about XSLT 3.0, because we don't have a feature for XSLT 3.0 yet.
    - In https://github.com/web-platform-tests/interop/issues/1137, `active-view-transition` was found for a proposal about View Transitions Types, because we don't have a feature for that yet.
    - In https://github.com/web-platform-tests/interop/issues/1113, `navigation` was found for a proposal about navigation precommit handlers, because we didn't have a feature for it yet (we do now: `navigation-precommit-handlers`).
  
  - Sometimes the author entered the wrong feature ID, because they couldn't find the right one:
    - In https://github.com/web-platform-tests/interop/issues/1082, the `reporting` feature was used instead of `permissions-policy` because the author couldn't find the right ID at the time.

In many of the above cases, someone (either the author or somebody with write access) added (or corrected) the feature ID in the opening comment after the issue had been filed. However, the script doesn't run again when that's the case. So we're stuck with the incorrect feature.

captainbrosset:
Fixes #1317.

* Refreshes individual proposals when issues are opened, edited, labeled, or reopened.
* Supports refreshing all open focus-area proposals manually or after updating the web-features dependency.
* Updates generated comments to use the latest explorer data, including developer signals, and more complete standards positions coverage.
* Handles renamed and split web-feature IDs.
* Some stricter validation and code cleanups.
* Upgrades web-features from 2.48.0 to 3.34.2.
* Adds some simple automated tests which we can improve later.
* Modernizes the workflow actions, permissions, dependency caching, etc.

Sorry that's a large PR with a lot of code changes. Hopefully, the summary above is helpful.

For what it's worth, the same code also lives at https://github.com/captainbrosset/interop-proposals-bot and I tested it by opening a few [issues](https://github.com/captainbrosset/interop-proposals-bot/issues) on that repo. If you want to test the new action/script more, feel free to open and edit issues on my personal repo.

### DIFF
--- a/.github/workflows/identify-web-features.yml
+++ b/.github/workflows/identify-web-features.yml
@@ -47,9 +47,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
         working-directory: scripts
-      - name: Run tests
-        run: npm test
-        working-directory: scripts
       - name: Refresh issue
         run: node identify-web-features.js --number ${{ github.event.issue.number }} --repo ${{ github.repository }}
         working-directory: scripts
@@ -74,9 +71,6 @@ jobs:
           cache-dependency-path: scripts/package-lock.json
       - name: Install dependencies
         run: npm ci
-        working-directory: scripts
-      - name: Run tests
-        run: npm test
         working-directory: scripts
       - name: Refresh all open proposals
         run: node identify-web-features.js --all-open-proposals --repo ${{ github.repository }}
--- a/.github/workflows/identify-web-features.yml
+++ b/.github/workflows/identify-web-features.yml
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     concurrency:
       group: identify-web-features-issue-${{ github.event.issue.number }}
-      cancel-in-progress: false
+      cancel-in-progress: true
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -62,7 +62,7 @@ jobs:
     runs-on: ubuntu-latest
     concurrency:
       group: identify-web-features-all-open-proposals
-      cancel-in-progress: false
+      cancel-in-progress: true
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/identify-web-features.yml
+++ b/.github/workflows/identify-web-features.yml
@@ -23,7 +23,14 @@ jobs:
 
   # This job runs when a single issue is opened, edited, etc.
   process-issue:
-    if: github.event_name == 'issues'
+    if: >-
+      github.event_name == 'issues' &&
+      (
+        github.event.action == 'opened' ||
+        github.event.action == 'edited' ||
+        github.event.action == 'reopened' ||
+        (github.event.action == 'labeled' && github.event.label.name == 'focus-area-proposal')
+      )
     runs-on: ubuntu-latest
     concurrency:
       group: identify-web-features-issue-${{ github.event.issue.number }}
--- a/.github/workflows/identify-web-features.yml
+++ b/.github/workflows/identify-web-features.yml
@@ -1,18 +1,19 @@
-# This GitHub Actions workflow identifies web-features in new and edited issues.
-# Identified features are posted as a comment on the issue.
+# This GitHub Actions workflow identifies web-features in new and edited focus-area-proposal issues.
+# Identified web-features are listed as a comment on the issue.
 name: Identify web-features in focus area proposals
 
 on:
-  # Trigger either when a single issue is opened or modified.
+  # Trigger the workflow either when a single issue is opened, edited, reopened, or labeled.
+  # The labeled trigger only applies when the label is 'focus-area-proposal'.
   issues:
     types: [opened, edited, labeled, reopened]
-  # Or when we update the package.json file (web-features dependency).
+  # Trigger the workflow when we update the package.json file (web-features dependency).
   push:
     branches: [main]
     paths:
       - scripts/package.json
       - scripts/package-lock.json
-  # Or manually.
+  # Lets us trigger the job manually.
   workflow_dispatch:
 
 permissions:
@@ -21,7 +22,7 @@ permissions:
 
 jobs:
 
-  # This job runs when a single issue is opened, edited, etc.
+  # The process-issue job runs when a single issue is opened, edited, reopened or labeled.
   process-issue:
     if: >-
       github.event_name == 'issues' &&
@@ -32,6 +33,8 @@ jobs:
         (github.event.action == 'labeled' && github.event.label.name == 'focus-area-proposal')
       )
     runs-on: ubuntu-latest
+    # Concurrency is used to ensure that only one job per issue runs at a time, with newer
+    # jobs canceling older ones.
     concurrency:
       group: identify-web-features-issue-${{ github.event.issue.number }}
       cancel-in-progress: true
@@ -53,10 +56,14 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-  # This job runs when the package.json file is changed (or manually), and updates all issues.
+  # The process-all-open-proposals job runs when the package.json file is changed, and updates all issues.
+  # This is used to refresh all open proposals when the web-features dependency is updated.
+  # The job can also be triggered manually.
   process-all-open-proposals:
     if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
+    # Concurrency is used to ensure that only one job per repository runs at a time, with newer
+    # jobs canceling older ones.
     concurrency:
       group: identify-web-features-all-open-proposals
       cancel-in-progress: true
--- a/.github/workflows/identify-web-features.yml
+++ b/.github/workflows/identify-web-features.yml
@@ -1,25 +1,78 @@
-# This GitHub Actions workflow identifies web features in new and edited issues.
+# This GitHub Actions workflow identifies web-features in new and edited issues.
 # Identified features are posted as a comment on the issue.
-name: Identify Web Features in Issues
+name: Identify web-features in focus area proposals
 
 on:
+  # Trigger either when a single issue is opened or modified.
   issues:
-    types: [opened, edited]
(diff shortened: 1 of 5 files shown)


.github/workflows/identify-web-features.yml
```diff
@@ -1,25 +1,78 @@
...
+# This GitHub Actions workflow identifies web-features in new and edited issues.
 # Identified features are posted as a comment on the issue.
-name: Identify Web Features in Issues
+name: Identify web-features in focus area proposals
 
 on:
+  # Trigger either when a single issue is opened or modified.
   issues:
-    types: [opened, edited]
+    types: [opened, edited, labeled, reopened]
```
foolip, line 8: Why labeled?

.github/workflows/identify-web-features.yml
```diff
@@ -1,25 +1,78 @@
...
+  issues: write
 
 jobs:
-  run-script:
+
+  # This job runs when a single issue is opened, edited, etc.
+  process-issue:
+    if: github.event_name == 'issues'
+    runs-on: ubuntu-latest
+    concurrency:
```
foolip, line 28: How many jobs might this trigger if we close issues en masse or add a lot of labels? GitHub Actions should just queue everything to eventually run, but if we know it'll be hundreds of jobs at the same time, that seems like it might break?

.github/workflows/identify-web-features.yml
```diff
@@ -1,25 +1,78 @@
...
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: npm
+          cache-dependency-path: scripts/package-lock.json
+      - name: Install dependencies
+        run: npm ci
+        working-directory: scripts
+      - name: Run tests
+        run: npm test
```
foolip, line 44: Do we really need to run the tests when updating issues?

.github/workflows/identify-web-features.yml
jgraham: I assume because we might add a label that identifies something as a proposal? But maybe we should also have an `if` condition that checks whether the issue actually has the label before running further? https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#example-using-an-object-filter has an example of a condition that filters on labels.

scripts/package.json
```diff
@@ -1,9 +1,12 @@
 {
   "name": "interop-scripts",
   "type": "module",
+  "scripts": {
+    "test": "node --test"
+  },
   "devDependencies": {
     "octokit": "^5.0.3",
-    "web-features": "^2.48.0",
+    "web-features": "^3.34.2",
```
jgraham, line 9: Should we also configure dependabot to automatically update this?

scripts/identify-web-features.js
```diff
@@ -185,156 +184,392 @@ function getDocsAsMarkdown(feature) {
...
+export async function postOrUpdateComment(octokit, repository, issueNumber, markdown) {
+  const existingComment = await removeDuplicateBotComments(
+    octokit,
+    repository,
+    await listBotComments(octokit, repository, issueNumber),
+  );
+
+  if (!existingComment) {
+    console.log(`Posting a new comment on issue #${issueNumber}...`);
+    const response = await octokit.request("POST /repos/{owner}/{repo}/issues/{issue_number}/comments", {
```
jgraham, line 374: Not sure why some GH requests are inline and others are functions.

scripts/identify-web-features.js
```diff
@@ -185,156 +184,392 @@ function getDocsAsMarkdown(feature) {
...
+      body: markdown,
+      headers: getGitHubHeaders(),
+    });
+
+    const commentToKeep = await removeDuplicateBotComments(
+      octokit,
+      repository,
+      await listBotComments(octokit, repository, issueNumber),
+    );
+    if (!commentToKeep) {
```
jgraham, line 386: Not sure I understand. If this is `undefined` then we didn't find any comments. It seems like that depends on us not finding the comment we just created? But surely that's not something we can reliably depend on?

scripts/identify-web-features.js
```diff
@@ -185,156 +184,392 @@ function getDocsAsMarkdown(feature) {
...
+      return "created";
+    }
+    if (commentToKeep.id !== response.data.id) {
+      console.log(`Another run created comment #${commentToKeep.id}; kept that comment instead.`);
```
jgraham, line 390: That's only true inside the `if` I think.

scripts/identify-web-features.js
```diff
@@ -185,156 +184,392 @@ function getDocsAsMarkdown(feature) {
...
+    headers: getGitHubHeaders(),
+  });
+  return "updated";
+}
+
+export async function processIssue(issue, {
+  octokit,
+  repository,
+  fetchImpl = fetch,
+  featureCatalog = webFeatures,
```
jgraham, line 425: I don't love this pattern where we're passing this through all the functions but also always providing a default. Let's just pass it through everywhere.

scripts/identify-web-features.js
```diff
@@ -185,156 +184,392 @@ function getDocsAsMarkdown(feature) {
...
+    console.log(`Skipping issue #${issue.number}: it is not an open ${REQUIRED_LABEL} issue.`);
+    return "skipped";
+  }
+
+  console.log(`Processing issue #${issue.number}: "${issue.title}"`);
+  const featureIds = findFeaturesInIssue(issue, featureCatalog);
+
+  // Handle moved and split features by redirecting to the target(s) in the catalog.
+  const processedFeatureIds = [];
+  for (const id of featureIds) {
```
jgraham, line 437: This loop could end up with duplicates; maybe we want `processedFeatureIds` to be a `Set` instead?

jgraham (commented):
Basically OK, although I'm quite confused with the comment creation flow.

In the future it would be really useful to separate out the changes into more commits; reviewing a mixture of minor formatting changes and multiple different feature changes is much harder.

.github/workflows/identify-web-features.yml
captainbrosset: Without the labeled trigger, an issue opened without the label and labeled later would never receive the generated web-features comment.
This is probably rare, but not inconceivable, right?

I can improve the `if` expression to avoid running further if the label isn't `focus-area-proposal`.

.github/workflows/identify-web-features.yml
captainbrosset: This is really here to avoid the case where the same issue is updated multiple times while a job for that issue is already running. With this concurrency block, the next jobs will be queued instead of running in parallel.

If I'm editing the opening comment of an issue twice in less time than it takes the job to run, then I don't run the 2 jobs at the same time, potentially ending up with the wrong content if the second ran faster than the first.

That said, I'm realizing that `cancel-in-progress:true` would make more sense. We only care about the final state of the issue, not the in-between edits. So a later job on an issue should be allowed to cancel an earlier job. I'll change this.

As for a batch editing of hundreds of issues, I'm not seeing an issue here. We'll let GitHub queue up all the jobs.

.github/workflows/identify-web-features.yml
captainbrosset: You're right, let me remove this.

scripts/package.json
captainbrosset: Yes we should. Not sure exactly how this works, but I can look into it next.

jgraham: https://github.com/web-platform-tests/interop/pull/1325 should be about right.

scripts/identify-web-features.js
captainbrosset: I was trying to cover (arguably rare) race conditions, and get confused. I think this if is not needed at all. If we create a comment, whether there was already one before or not, is enough of a guarantee that one will exist now.

captainbrosset: Removed.

scripts/identify-web-features.js
captainbrosset: I don't understand this comment. Another comment could have been added in the meantime, and here we keep that one and update it if needed.

captainbrosset:
> In the future it would be really useful to separate out the changes into more commits

Apologies for the messy initial changes. I've addressed review comments in a series of logical commits now.

.github/workflows/identify-web-features.yml
```diff
@@ -1,25 +1,86 @@
...
+  contents: read
+  issues: write
 
 jobs:
-  run-script:
+
+  # The process-issue job runs when a single issue is opened, edited, reopened or labeled.
+  process-issue:
+    if: >-
+      github.event_name == 'issues' &&
```
jgraham, line 28: This is fine, but one could also use a condition like `contains(github.event.issue.labels.*.name, 'focus-area-proposal')` to only select focus area proposal issues irrespective of the trigger.

jgraham (approved):

